PR #9, opened 2021-10-05
Change the language name to plrustu 

jim-mlodgenski:
The PostgreSQL uses the convention of suffixing the name with a "u" to indicate an untrusted language. This adds that convention and opens the possibility of adding "trusted" plrust in the future.


eeeebbbbrrrr:
We've long since moved past this being the correct approach.  When we get postgrestd working on MacOS I think then we can take a look at making a proper distinction between trusted and untrusted.  Until then, having that knowledge compiled into the plrust binary is good enough.

PR #19, opened 2022-03-16
Respect check_function_bodies

Hoverbear:
This is a setting on by default in PostgreSQL: https://postgresqlco.nf/doc/en/param/check_function_bodies/

This was code @eeeebbbbrrrr commented out but seems to be working right.


eeeebbbbrrrr:
I don't think this is correct.  If anything, I think that `if` block should be removed entirely.

I can't remember why I commented out that line (I generally don't commit commented-out  code for that very reason), but I suspect I was just playing around with if it was necessary or not, and it wasn't.

the `validator` function for a pl language handler is supposed to syntax check (or in our case, compile) the function, not "load it".  

We ignore the `check_function_bodies` GUC for the reasons described in comments at the top of that function.

Hoverbear:
Ok! I'll update this PR to reflect that removal instead.

Hoverbear:
@eeeebbbbrrrr could you take another look?

PR #29, opened 2022-03-25
Repair cfg selection on x86 Mac

Hoverbear:
We have some workaround in x86 Mac which we weren't correctly feature flagging in #12 .


eeeebbbbrrrr (approved):
LGTM!

PR #30, opened 2022-03-26
update copyright headers

eeeebbbbrrrr:


Hoverbear:
I left some feedback in https://github.com/tcdi/pgx/pull/511 which applies here as well.

.licensure.yml
```diff
@@ -0,0 +1,57 @@
...
+  - ".*\\.md"
+  - ".*\\.nix"
+  - ".*\\.yml"
+licenses:
+  - files: any
+    ident: PostgreSQL
+    authors:
+      - name: ZomboDB, LLC and Technology Concepts & Design, Inc.
+        email: support@tcdi.com
+    template: |
```
Hoverbear, line 24: The PostgreSQL license does not require copyright headers in the original or derivative works. This could be removed.

.licensure.yml
```diff
@@ -0,0 +1,57 @@
...
+      Portions Copyright 2020-2021 ZomboDB, LLC. 
+      Portions Copyright 2021-[year] Technology Concepts & Design, Inc. <support@tcdi.com>.
+      All rights reserved.
```
Hoverbear, line 27: This term applied per-source-file seems to differ from the license in the root of the repo, creating an impression that these files follow a more restrictive license ("All rights reserved"). It should be updated in the root license, if this is included at all.

Issue #28, opened 2022-03-23
RFC: Artifact storage options


Hoverbear:
Right now PL/Rust keeps produced shared objects on the local filesystem in the directory specified by the `plrust.work_dir`  GUC.

We'd like to have an option for PL/Rust to keep these artifacts in a specific system catalog (configurable via GUC).

This system catalog would store the artifacts as a `bytea` in a table like:

```sql
CREATE TABLE plrust.artifacts (
    name text,
    data bytea,
); 
```

The setting could be like `plrust.artifact_store`  and it could support `directory` or `table`.

If `directory` is specified, we could have a GUC `plrust.artifact_directory` (replacing `plrust.work_dir`).

If `table` is specified, we could have a GUC `plrust.artifact_table`. (taking a string)

We should also consider what the migration story looks like between them, and how PL/Rust does (or does not) facilitate that.

JohnHVancouver:
I'm curious how the workflow of this would work if we stored it in the plrust.artifacts table. 
When the user calls create function with PL/Rust, then:
1. We setup the cargo create the same way and get a `.so`
2. Store the `.so` as a bytea into `plrust.artifacts`

Then when they execute the function, would we need to:
1. Query plrust.artifacts to get the data
2. Write the data as `.so` onto disk and then call it?

@eeeebbbbrrrr  mentioned one of the benefits would be to avoid recompiling the code on replicas of the same architecture, which makes sense.

Is there some mechanism to optimize the look-up/write-to-disk away? I guess you could do the look-up for each PL/Rust function on a per-transaction basis...

Even with wasm, assuming that's the direction we're moving towards and we store the wasm in `pg_proc.prosrc` as suggested by Eric, I think you still need to write the `.wasm ` out to disk to be added as a module in order to load/execute it?
We could store it as a `.wat` instead and avoid the write-to-disk and having to handle race conditions/formatting there but that doesn't seem to perform as nicely.

I did a quick benchmark where I had a plrust.function executing a `.wasm`, vs a `.wat` on the filesystem that I converted it to with wasm2wat (https://github.com/WebAssembly/wabt). My PL/Rust function just returned a number and I got around 0.194ms with `.wasm` compared to 0.265ms with `.wat` 

```
cargo pgx stop
cargo pgx start
cargo pgx connect
cat /dev/null > 14.log

set log_duration = 'on'
select return_int(); \watch 0.0001

[...]

PR #33, opened 2022-04-06
Add SQL functions to cleanup PL/Rust scratch space

JohnHVancouver:
```
When a function is dropped, the corresponding files
in the scratch directory are not cleaned up. Since
PL/Rust could be used in environments where the users
do not have access to the filesystem or superuser, have
it provide mechanisms to cleanup stale files in the
scratch directory.
```

Creating this pull request to get the conversation started whether or not this is useful. I still need to do some testing, and there's probably a lot of the Rust that can be simplified into more rust-isms (that I need to learn...)

There's some cleaning up that still needs to be done, mainly this still exposes metadata information about other directory but basic sanity tests seem to pass, see testing section.

I also wasn't too sure how to actually write tests for this within PL/Rust itself, any suggestions on this part?

The initial goal of this was to create enough wrapper functions, mainly `plrust.remove_fn_from_work_dir`  so that it could be dropped automatically via event triggers. I had an idea that on `DROP FUNCTION` we could clean up the corresponding scratch space on `sql_drop` [0]

It doesn't seem as nice as I would want it to be since the event [1] only contains information about the schema name, and the fn_oid. There isn't information if it's a PL/Rust function or not. Although I suppose it doesn't hurt to create an EVENT TRIGGER that on every drop function just checks if there's a corresponding format in the scratch space and remove accordingly, which sounds like it'd work?

If the user was superuser (pg_ls_dir + COPY TO PROGRAM 'rm /scratch/foo' this isn't necessary but these aren't necessarily available to all users.

**Testing**

```
plrust=# select plrust.list_function_in_work_dir();
                     list_function_in_work_dir                     
-------------------------------------------------------------------
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40980_0.so
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_57497_1.so
 /Users/hsuchen/git/zombodb/plrust/scratch/.rustc_info.json
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40980_1.so
 /Users/hsuchen/git/zombodb/plrust/scratch/plrust_interface
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_57497_0.so
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40979
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40977
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_0_40980
(9 rows)

-- test remove directory
plrust=# set session authorization testuser;
SET
plrust=> select plrust.remove_fn_from_work_dir_priviliged('fn16384_0_40980');
ERROR:  Only superusers are allowed to remove arbitrary files
CONTEXT:  src/plrust.rs:246:9
plrust=> reset session authorization;
RESET
plrust=# select plrust.remove_fn_from_work_dir_priviliged('fn16384_0_40980');
 remove_fn_from_work_dir_priviliged 
------------------------------------
 
(1 row)

plrust=# select plrust.list_function_in_work_dir();                                                                                                                                                    list_function_in_work_dir                     
-------------------------------------------------------------------
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40980_0.so
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_57497_1.so
 /Users/hsuchen/git/zombodb/plrust/scratch/.rustc_info.json
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40980_1.so
 /Users/hsuchen/git/zombodb/plrust/scratch/plrust_interface
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_57497_0.so
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40979
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40977
(8 rows)

-- test remove file
plrust=# select plrust.remove_fn_from_work_dir_priviliged('fn16384_2200_57497_0.so');
 remove_fn_from_work_dir_priviliged 
------------------------------------
 
(1 row)

plrust=# select count(*) from plrust.list_function_in_work_dir();
 count 
-------
     7
(1 row)

-- test remove file that is somewhere else TODO: fix leaks information about files 
plrust=# select plrust.remove_fn_from_work_dir_priviliged('../fn16384_2200_57497_0.so');
ERROR:  called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
CONTEXT:  src/plrust.rs:249:70
plrust=# select plrust.remove_fn_from_work_dir_priviliged('../test');
ERROR:  Unable to remove a file not in the work directory
CONTEXT:  src/plrust.rs:252:9

plrust=# select plrust.list_function_in_work_dir();
                     list_function_in_work_dir                     
-------------------------------------------------------------------
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40980_0.so
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_57497_1.so
 /Users/hsuchen/git/zombodb/plrust/scratch/.rustc_info.json
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40980_1.so
 /Users/hsuchen/git/zombodb/plrust/scratch/plrust_interface
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40979
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40977
(7 rows)

plrust=# select plrust.remove_fn_from_work_dir(40980, 'public');
 remove_fn_from_work_dir 
-------------------------
 
(1 row)

plrust=# select plrust.list_function_in_work_dir();
                     list_function_in_work_dir                     
-------------------------------------------------------------------
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_57497_1.so
 /Users/hsuchen/git/zombodb/plrust/scratch/.rustc_info.json
 /Users/hsuchen/git/zombodb/plrust/scratch/plrust_interface
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40979
 /Users/hsuchen/git/zombodb/plrust/scratch/fn16384_2200_40977
(5 rows)
```

-------
[0] https://www.postgresql.org/docs/current/event-trigger-definition.html
[1] https://www.postgresql.org/docs/current/functions-event-triggers.html#PG-EVENT-TRIGGER-SQL-DROP-FUNCTIONS

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,6 +76,24 @@ fn recompile_function(
     }
 }
 
+#[pg_extern]
+fn list_function_in_work_dir() -> impl std::iter::Iterator<Item = String> {
+    plrust::list_functions()
+}
+
+#[pg_extern]
+fn remove_fn_from_work_dir(fn_oid: pg_sys::Oid, schema_name: String) {
+    let removed = plrust::remove_fn_from_work_dir(fn_oid, schema_name);
+    if !removed {
+        panic!("No file was removed")
+    }
+}
+
+#[pg_extern]
+fn remove_fn_from_work_dir_priviliged(basename: String) {
+    plrust::remove_file_from_work_dir(basename);
+}
+
 extension_sql!(
     r#"
 CREATE LANGUAGE plrust
--- a/src/plrust.rs
+++ b/src/plrust.rs
@@ -157,6 +157,121 @@ pub(crate) unsafe fn lookup_function(
     }
 }
 
+pub(crate) fn list_functions() -> impl std::iter::Iterator<Item = String> {
+    let paths = std::fs::read_dir(gucs::work_dir()).unwrap();
+    let mut valid_fns = Vec::new();
+
+    // Provide a santized version of pg_ls_dir but only for the work_dir
+    // Useful when PL/Rust is being provided to an unpriviliged user
+    let superuser = unsafe { pg_sys::superuser() };
+    let db_oid = unsafe { pg_sys::MyDatabaseId };
+
+    for entry in paths {
+        let path = String::from(entry.unwrap().path().to_string_lossy());
+
+        // If superuser, provide a view of the contents of the directory
+        // Otherwise provide only a view of files containing the oid of the database.
+        // This is to prevent metadata information from being leaked in multi-tenant systems
+        if superuser {
+            valid_fns.push(path);
+        } else {
+            let sanitized_path = String::from(
+                std::path::Path::new(&gucs::work_dir())
+                    .join(format!("{}{}", "fn", db_oid))
+                    .to_string_lossy(),
+            );
+            if path.starts_with(&sanitized_path) {
+                valid_fns.push(path);
+            }
+        }
+    }
+
+    valid_fns.into_iter()
+}
+
+pub(crate) fn remove_fn_from_work_dir(fn_oid: pg_sys::Oid, schema_name: String) -> bool {
+    let mut removed = false;
+    let schema = std::ffi::CString::new(schema_name).unwrap();
+
+    // Don't need to sanitize schema as it gets converted into an Oid
+    let ns_oid = unsafe { pg_sys::get_namespace_oid(schema.as_ptr(), false) };
+    let db_oid = unsafe { pg_sys::MyDatabaseId };
+
+    let crate_name = crate_name_fmt(db_oid, ns_oid, fn_oid);
+
+    let work_dir = gucs::work_dir();
+    for entry in work_dir
+        .read_dir()
+        .expect("Unable to iterate through PL/Rust work directory")
+    {
+        if let Ok(entry) = entry {
+            let canonical_path = entry.path().canonicalize().unwrap();
+
+            // Check that the canonical path of the entry is in the work directory
+            // and the file_name/basename lines up with our crate format
+            // We check for is_dir and is_file to protect against symlink but since the
+            // operations are not atomic it's still possible for us to accidentally delete
+            // a file in an unintended location.
+            if canonical_path.starts_with(&work_dir)
+                && canonical_path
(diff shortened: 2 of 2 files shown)


eeeebbbbrrrr:
Hey @JohnHVancouver! Thanks for the PR.

I don't think this is something plrust should expose to users.  We have an open issue around how we *should* be storing artifacts (https://github.com/tcdi/plrust/issues/28), and I think the general consensus is that the filesystem is not actually the correct place.

If it were, however, to be the correct place, the proper way of handling this would be to hook Postgres' "process utility hook" (which pgx allows), intercept `DROP FUNCTION` statements, figure out if the function is a plrust function by groveling through the statement's parse tree, and queue up the associated `.so` to be deleted only when the transaction commits.  This way we're consistent with Postgres' MVCC rules, place nice with transactions, and properly support WAL replication to standby servers.

In fact, we'll likely need to have that kind of code in place anyways, even if we instead start storing artifacts in a table in the database (as #28) suggests.

Alternatively, if we continue to move forward with WASM, since that artifact is text, we'd opt to store it in the `prosrc` column of the `pg_proc` catalog, along side the function's Rust source.

What are your thoughts?

JohnHVancouver:
Thanks @eeeebbbbrrrr 
I agree that, if we did want to clean it up on a transactional basis it'd be a lot cleaner to hook into `process_utility_hook`, I need to explore that more (although I wonder how sub-transactions work but doubt it's worth worrying about).

#28 is interesting and does sort of address the problem. I'll comment on that instead.

PR #35, opened 2022-04-13
Wasm execution wit

Hoverbear:
Similar to #34 , this builds atop it to include WIT interface types.


src/interface.rs
```diff
@@ -0,0 +1,23 @@
...
+wit_bindgen_wasmtime::import!("src/guest.wit");
+
+#[derive(Default)]
+pub struct Host;
+
+impl host::Host for Host {
+    fn get_one_with_args(
+        &mut self,
+        query: &str,
+        args: Vec<host::ValueParam<'_>>
```
JohnHVancouver: Where does `ValueParam` and `ValueResult` get defined from? Is it accessible as part of the

PR #43, opened 2022-05-04
"post-alloc" skeleton

workingjubilee:
This doesn't include out-of-line crates I was fooling around with that are referenced here, as I wasn't sure what the best restructuring of things would be at the time, so I just pushed them somewhere else. At the moment I have just been using more direct FFI calls to Postgres for the allocator's implementation, but I imagine with some refactoring of PGX, a possible more sophisticated implementation might mesh itself more seamlessly with existing code in PGX. Or not.

One of the unresolved questions is "wait, why do tests randomly fail if I use `palloc` instead of `SPI_palloc`?"

### DIFF
--- a/src/plrust.rs
+++ b/src/plrust.rs
@@ -231,7 +231,7 @@ fn create_function_crate(fn_oid: pg_sys::Oid, crate_dir: &PathBuf, crate_name: &
     std::fs::write(
         &cargo_toml,
         &format!(
-                r#"[package]
+            r#"[package]
 name = "{crate_name}"
 version = "0.0.0"
 edition = "2021"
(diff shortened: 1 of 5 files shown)


workingjubilee:
This is likely blocked on https://github.com/tcdi/pgx/pull/562 and possibly other things, like finding a better place for the allocator code, inside a proper crate perhaps.

src/plrust.rs
```diff
@@ -258,15 +259,21 @@ crate-type = ["cdylib"]
...
+{experimental_deps}
 
 [profile.release]
+debug-assertions = true
 panic = "unwind"
 opt-level = 3
 lto = "fat"
 codegen-units = 1
 "#,
+            experimental_deps = match std::env::var("experimental_crates") {
```
Hoverbear, line 273:
```suggestion
            experimental_deps = match std::env::var("PLRUST_EXPERIMENTAL_CRATES") {
```

Hoverbear (approved):
Rest seems good!

PR #45, opened 2022-05-16
Improved Error handling

Hoverbear:
Improves error handling code in PL/Rust.

Includes:

* `thiserror`
* `eyre`/`tracing` wired up to not interfere with pgx
* Spantraces and error sections supported
* *Most* unwraps turned into errors, this is just a first pass.
* `tracing_level` GUC for customizing the log level of PL/Rust itself
* `#[pg_extern]` functions now wrap functions which return errors, and handle them.
* Updates flake for CI

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -14,7 +14,7 @@ pub enum PlRustError {
     LibLoading(#[from] libloading::Error),
     #[cfg(all(target_os = "macos", target_arch = "x86_64"))]
     #[error("Generation error (Mac OS x86_64 specific): {0}")]
-    Generation(#[from] crate::generation::Error),
+    Generation(#[from] crate::plrust::generation::Error),
     #[error("Creating crate directory in plrust.work_dir GUC location: {0}")]
     CrateDirectory(std::io::Error),
     #[error("Executing `cargo build`: {0}")]
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,4 +1,3 @@
-
 #[derive(thiserror::Error, Debug)]
 pub enum PlRustError {
     #[error("Failed pg_sys::CheckFunctionValidatorAccess")]
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,25 @@
+
+#[derive(thiserror::Error, Debug)]
+pub enum PlRustError {
+    #[error("Failed pg_sys::CheckFunctionValidatorAccess")]
+    CheckFunctionValidatorAccess,
+    #[error("pgx::pg_sys::FunctionCallInfo was Null")]
+    NullFunctionCallInfo,
+    #[error("pgx::pg_sys::FmgrInfo was Null")]
+    NullFmgrInfo,
+    #[error("libloading error: {0}")]
+    LibLoading(#[from] libloading::Error),
+    #[cfg(all(target_os = "macos", target_arch = "x86_64"))]
+    #[error("Generation error (Mac OS x86_64 specific): {0}")]
+    Generation(#[from] crate::generation::Error),
+    #[error("Creating crate directory in plrust.work_dir GUC location: {0}")]
+    CrateDirectory(std::io::Error),
+    #[error("Executing `cargo build`: {0}")]
+    CargoBuildExec(std::io::Error),
+    #[error("`cargo build` failed: {0}")]
+    CargoBuildFail(String),
+    #[error("Produced shared object not found")]
+    SharedObjectNotFound,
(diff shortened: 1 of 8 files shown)


src/error.rs
```diff
@@ -0,0 +1,38 @@
+#[derive(thiserror::Error, Debug)]
+pub enum PlRustError {
```
Hoverbear, line 2: This should be split up later once the crate has more structure, but it's a start.

src/gucs.rs
```diff
@@ -30,6 +31,14 @@ pub(crate) fn init() {
         &PLRUST_PG_CONFIG,
         GucContext::Sighup,
     );
+
+    GucRegistry::define_string_guc(
+        "plrust.tracing_level",
+        "The tracing level to use while running pl/rust",
+        "The tracing level to use while running pl/rust. Should be `error`, `warn`, `info`, `debug`, or `trace`",
+        &PLRUST_TRACING_LEVEL,
```
Hoverbear, line 39: We could, alternatively, allow the user to pass a set of tracing directives.

src/lib.rs
```diff
@@ -65,14 +142,22 @@ fn recompile_function(
     fn_oid: pg_sys::Oid,
 ) -> (
     name!(library_path, Option<String>),
-    name!(cargo_output, String),
+    name!(stdout, Option<String>),
+    name!(stderr, Option<String>),
+    name!(plrust_error, Option<String>),
 ) {
+    tracing::error!("Swoop de woop");
```
Hoverbear, line 149: omg...

Hoverbear:
Ok this is ready for review now. 😅

src/gucs.rs
workingjubilee: Neither jumps out at me as obviously preferable so this is probably fine.

src/logging.rs
```diff
@@ -0,0 +1,61 @@
+use std::io::Write;
+
+pub(crate) struct PgxGuestWriter<const IS_STDERR: bool>;
+
+impl<const IS_STDERR: bool> Write for PgxGuestWriter<IS_STDERR> {
+    fn write(&mut self, data: &[u8]) -> std::result::Result<usize, std::io::Error> {
+        let content = std::str::from_utf8(data).expect("Could not interpret stdout as UTF-8");
```
workingjubilee, line 7: Hmm. I am not following how non-UTF8 data would get into our stdout/stderr yet...? Is there no way we can ban that on the type level?

src/plrust.rs
```diff
@@ -184,18 +197,16 @@ pub(crate) fn compile_function(fn_oid: pg_sys::Oid) -> Result<(PathBuf, String),
...
-        .expect("failed to build function shared library");
+        .map_err(PlRustError::CargoBuildExec)?;
 
-    let mut output_string = String::new();
-    unsafe {
-        output_string.push_str(&String::from_utf8_unchecked(cargo_output.stdout));
-        output_string.push_str(&String::from_utf8_unchecked(cargo_output.stderr));
-    }
+    let stdout = String::from_utf8(cargo_output.stdout).map_err(PlRustError::CargoOutputNotUtf8)?;
+    let stderr = String::from_utf8(cargo_output.stderr).map_err(PlRustError::CargoOutputNotUtf8)?;
```
workingjubilee, line 203: I'm pretty sure it will be a cold day in hell before these return anything but, however, this is still better.

src/plrust.rs
```diff
@@ -344,19 +367,23 @@ fn plrust_fn_{fn_oid}"#
...
+    (
+        pg_sys::Oid,
+        String,
+        String,
+        Vec<(PgOid, Option<String>)>,
+        (PgOid, bool),
+        bool,
+    ),
+    PlRustError,
+> {
```
workingjubilee, line 386: I begin to resent this large return tuple. Its girth mocks me. It was already like that, however.

workingjubilee (commented):
This looks much nicer and I am sure will prove much more useful soon, I only have one trivial question.

src/logging.rs
workingjubilee: Apparently "theoretically a build log could". Eh, I suppose.

Hoverbear: In theory a build.rs script error log could dumb non-UTF-8 data to stdout but that's highly irregular.

workingjubilee (approved):

src/plrust.rs
Hoverbear: That's my next task after #46, I was hoping to give you a little idea we could find consensus over first though!

PR #46, opened 2022-05-16
Syn and quote

Hoverbear:
Switches us entirely to syn/quote and toml types and handling to improve both our defensive posture as well as improve the future maintainability of the code.

It also does some various tunings to how we handle errors, refining what #45 did.

Adds a few tests I was using during my testing.


src/plrust.rs
```diff
@@ -478,7 +573,59 @@ fn parse_source_and_deps(code: &str) -> (String, String) {
...
+        toml::from_str(&deps_block).map_err(PlRustError::ParsingDependenciesBlock)?;
+
+    let user_code: syn::Block =
+        syn::parse_str(&format!("{{ {code_block} }}")).map_err(PlRustError::ParsingCodeBlock)?;
+
+    Ok((user_code, user_dependencies))
+}
+
+#[tracing::instrument(level = "debug", skip_all, fields(type_oid = type_oid.value()))]
+fn oid_to_syn_type(type_oid: &PgOid) -> Result<syn::Type, PlRustError> {
```
Hoverbear, line 586: I ripped this from wasm code and it doesn't account for `is_owned`

src/plrust.rs
```diff
@@ -231,62 +263,135 @@ pub(crate) fn compile_function(fn_oid: pg_sys::Oid) -> eyre::Result<(PathBuf, St
...
+                        crates_io_table.insert("pgx".into(), toml::Value::Table(pgx_table));
+                        entry.or_insert(toml::Value::Table(crates_io_table));
+                    }
+                    _ => {
+                        return Err(PlRustError::GeneratingCargoToml).wrap_err(
+                            "Setting `[patch]`, already existed (and wasn't expected to)",
+                        )?
+                    }
+                },
+            };
```
workingjubilee, line 386: Need to go down a level here.

```suggestion
            match std::env::var("PLRUST_EXPERIMENTAL_CRATES") {
                Err(_) => (),
                Ok(path) => {
                    match cargo_manifest
                        .entry("patch")
                        .or_insert(toml::Value::Table(Default::default()))
                        .as_table_mut()
                        .unwrap() // infallible
                        .entry("crates-io")
                    {
                        entry @ toml::value::Entry::Vacant(_) => {
                            let mut pgx_table = toml::value::Table::new();
                            pgx_table.insert("path".into(), toml::Value::String(path.to_string()));
                            let mut crates_io_table = toml::value::Table::new();
                            crates_io_table.insert("pgx".into(), toml::Value::Table(pgx_table));
                            entry.or_insert(toml::Value::Table(crates_io_table));
                        }
                        _ => {
                            return Err(PlRustError::GeneratingCargoToml).wrap_err(
                                "Setting `[patch]`, already existed (and wasn't expected to)",
                            )?
                        }
                    }
                }
            };
```

Hoverbear: Oh no! Thank you

PR #52, opened 2022-06-06
Different CI

Hoverbear:
Use not-Nix so we can have not-sandboxed tests so we can test random user crates etc easier.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,76 +1,155 @@
-on: [push]
+name: PL/Rust
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+env:
+  RUSTFLAGS: -Copt-level=0
+  RUST_BACKTRACE: 1
+  CARGO_INCREMENTAL: "false"
+  CARGO_PGX_VERSION: "0.4.5"
+  # CARGO_LOG: cargo::core::compiler::fingerprint=info # Uncomment this to output compiler fingerprint info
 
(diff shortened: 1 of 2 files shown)


Hoverbear:
Hey @BradyBonnette would you be kind enough to figure out if I did an ok job porting your ci script?

.github/workflows/ci.yml
```diff
@@ -1,76 +1,155 @@
...
+      uses: actions/cache@v2
+      continue-on-error: false
+      with:
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: v0-${{ runner.os }}-${{ hashFiles('**/Cargo.lock', '.github/workflows/tests.yml') }}
```
BradyBonnette, line 128: @Hoverbear 

Will `v0` here in the cache key need to change in the future?

BradyBonnette (commented):
Everything looks good, though I had a question regarding one tiny thing (see inline comment regarding cache key)

.github/workflows/ci.yml
Hoverbear: If we change the cache layout, yes! That's its purpose!

BradyBonnette: Ah, I see. Works for me :)

BradyBonnette (approved):

workingjubilee:
Seems good to me~!

PR #56, opened 2022-06-28
Enable using postgrestd

workingjubilee:
This lays out some target selection logic using a `PLRUST_TARGET` env var, adds a feature for selecting an appropriate tuple by default, and otherwise provides for actually using postgrestd. This is logically complete but doesn't actually include the full test and build script. I'm going to let CI/review make noises at me before I do anything further that would mangle that signal.


workingjubilee:
( sorry, last commit was erroneous. )

workingjubilee:
Link error against undefined symbols, but... only on macOS, apparently, as Ubuntu does fine? Okay then.

src/user_crate/target.rs
```diff
@@ -0,0 +1,80 @@
+/*! PL/Rust adopts the tactic of always explicitly specifying which target to build for.
+    This prevents using the "fallback" logic of Cargo leaving builds in an unlabeled directory.
+    This is a precaution as PL/Rust is a cross-compiler.
+    so a normal build-and-test cycle may create artifacts for multiple targets.
+!*/
```
Hoverbear, line 5: I think the `!` is unnecessary but I ain't mad if it stays!

src/user_crate/target.rs
```diff
@@ -0,0 +1,80 @@
...
+        if t != &"" {
+            s.push('-');
+            s.push_str(t);
+        }
+    }
+    s
+}
+
+#[derive(Debug)]
+#[allow(dead_code)] // Such is the life of cfg code
```
Hoverbear, line 50: The rest of the code uses `thiserror` (see `src/error.rs`, would you be OK using that here as well?

Hoverbear:
Really excited to give this a spin. Is there any parts of postgrestd or the other libraries you'd like me to review as well?

src/user_crate/target.rs
workingjubilee: It feels nice and symmetric? :shrug:

src/user_crate/target.rs
workingjubilee: Yes, that would be fine!

Cargo.toml
```diff
@@ -55,3 +57,10 @@ panic = "unwind"
 opt-level = 3
 lto = "fat"
 codegen-units = 1
+
+[patch.crates-io]
+libc = { git = "https://github.com/workingjubilee/libc", branch = "postgres-os" }
+getrandom = { git = "https://github.com/workingjubilee/getrandom", branch = "postgres-os" }
+ring = { git = "https://github.com/workingjubilee/ring", branch = "postgres-os" }
```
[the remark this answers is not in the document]
workingjubilee: Yeah, that won't happen immediately here, but Soon.

.github/workflows/ci.yml
```diff
@@ -128,10 +127,10 @@ jobs:
 
     - name: Install cargo-pgx
       continue-on-error: true # We might have it.
-      run: cargo install --force cargo-pgx --version $CARGO_PGX_VERSION
```
Hoverbear: Is it possible we need to install the `cargo-pgx` off the branch?

Hoverbear:
Hmmm so does this mean `#[no_std]` in `src/user_crate/skeleton.rs` is not required?

workingjubilee:
Yeah, I thiiink the `#![no_std]` annotation is no longer required.

workingjubilee:
Okay, I am throwing in the towel on making Intel Mac work. One is in the mail to me so I will be able to plug it in and poke around and fix this soon, but for right now, eff it.

Hoverbear:
@workingjubilee I'm fine with not doing this on intel Macs yet.  They can just not use that feature. It is a dying platform anyways.

Hoverbear (approved):

Hoverbear:
Both #51 and #57 work perfectly fine on top this branch, good job. :)

PR #58, opened 2022-07-11
Use pallocator as a crate

workingjubilee:
Finally refactor out the allocator code to make the code look like real, organized crates instead. The "skeleton" is now just a line of `use ::pgx::*;` so as to not break `#[pg_extern]`. Also serves as a demonstration the PR branch it uses for the pallocator crate is viable.

### DIFF
--- a/src/user_crate/state_generated.rs
+++ b/src/user_crate/state_generated.rs
@@ -219,7 +219,7 @@ impl StateGenerated {
 
                     [dependencies]
                     pgx = { version = "0.5.0-beta.0", features = ["postgrestd"], git = "https://github.com/tcdi/pgx", branch = "develop" }
-                    pallocator = { version = "0.1.0", git = "https://github.com/tcdi/postgrestd", branch = "fix-pallocator" }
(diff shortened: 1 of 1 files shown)


src/user_crate/state_generated.rs
```diff
@@ -126,8 +126,7 @@ impl StateGenerated {
 
     #[tracing::instrument(level = "debug", skip_all, fields(fn_oid = %self.fn_oid))]
     pub(crate) fn lib_rs(&self) -> eyre::Result<syn::File> {
-        let mut skeleton: syn::File =
-            syn::parse_str(include_str!("./skeleton.rs")).wrap_err("Parsing skeleton code")?;
+        let mut skeleton: syn::File = syn::parse_str("use ::pgx::*;").wrap_err("Parsing skeleton code")?;
```
Hoverbear, line 129: I think this won't fail so you can do this instead I think:

```suggestion
        let mut skeleton: syn::File = syn::parse_quote!(use ::pgx::*;");
```

Hoverbear (approved):
Rest looks good!

PR #60, opened 2022-07-21
Adds postgresd/ directory to .gitignore, it should not ever be committed

BradyBonnette:


workingjubilee (approved):
That is quite correct! Thank you!

PR #61, opened 2022-07-25
Explain postgrestd in README

workingjubilee:


README.md
```diff
@@ -38,6 +38,23 @@ Option | Description
...
+and so is not currently set by default.
+
+This initial build process requires a normal installation of Rust via [`rustup`](https://rustup.rs)
+and for the relevant location to be writeable on the building host.
+
+```bash
+cd plrust
+./build
+PLRUST_TARGET="x86_64-unknown-linux-postgres" cargo build
+```
```
BradyBonnette, line 57: @workingjubilee 

One of the things I had to run was something such as:

`cargo pgx run pg14 --features target_postgrestd`

Does that belong in this section?

workingjubilee: That's the alternative to setting `PLRUST_TARGET`. We can just enable that though, also.

BradyBonnette:
@workingjubilee uh oh:  https://github.com/tcdi/plrust/runs/7511647832?check_suite_focus=true

workingjubilee:
So I did some work on analyzing this problem and clawed together a few hacks, but it has to do with pgx-tests not being prepared for this use-case fully. It doesn't inherit `--no-default-features` properly. Any solution working around this is a hack and none provide significantly improved UX, in my opinion, so I am going to simply revert that.

BradyBonnette (approved):

PR #66, opened 2022-08-12
Test panics don't kill the database

### LINKED ISSUE #64: Test that various panics can't abort the database
Normal Rust control flow should always simply gracefully shutdown. We should test to make sure this happens. Now that https://github.com/tcdi/pgx/pull/613 has landed we can actually do this.

We aren't going to worry about the specific case of someone deliberately trying to signal an abort, though. That's on them.

workingjubilee:
This should close #64.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -291,6 +291,87 @@ mod tests {
             Some("operation not supported on this platform".to_string())
         );
     }
+
+    #[pg_test]
+    #[search_path(@extschema@)]
+    #[should_panic]
+    fn pgx_can_panic() {
+        panic!()
+    }
+
+    #[pg_test]
+    #[search_path(@extschema@)]
+    #[should_panic]
+    fn plrust_can_panic() {
+        let definition = r#"
+            CREATE FUNCTION shut_up_and_explode()
+            RETURNS text AS
+            $$
+                panic!();
+                None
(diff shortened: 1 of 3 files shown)


src/tests.rs
```diff
@@ -291,6 +291,87 @@ mod tests {
...
+        Spi::run(definition);
+
+        let retval = Spi::get_one::<String>("SELECT say_hello();\n");
+        assert_eq!(retval, Some("Hello world\n".into()));
+    }
+
+    /// This test is... meta. It's intended to sleep enough to usually go last.
+    /// If a previous test panics in a way so severe it aborts Postgres?
+    /// This test will fail as a result.
+    /// Ideally this should be slow but still sometimes finish second-to-last.
```
BradyBonnette, line 345: Out of curiosity, what happens if this test does finish second-to-last?

workingjubilee: Nothing, but if it never finishes not-last, it's too slow (my initial draft had it add about a minute to the full test suite on my machine, a test suite which only takes about 2 minutes locally).

BradyBonnette: Ah ok, thanks!

BradyBonnette (approved):
Ran it locally as well. Looks good! Still laughing at some of those function names.

PR #71, opened 2022-08-18
Explain the `PLRUST_TARGET` env var better

workingjubilee:


README.md
```diff
@@ -45,7 +45,11 @@ Doing so places a copy of the necessary libraries used by Rust for `std` into th
 which is the location that rustc will look for building those libraries.
 
 Further usage requires setting the `PLRUST_TARGET` to the appropriate tuple, as this is still an experimental feature,
-and so is not currently set by default. Passing `--features target_postgrestd` to `cargo` also works.
+and so is not currently set by default. Passing `--features target_postgrestd` to `cargo` also works. These are subtly different:
+
+- `PLRUST_TARGET` is evaluated when the PL/Rust handler builds a function
```
BradyBonnette, line 50: Does this mean the plrust handler sets this `PLRUST_TARGET` environment variable when it goes to build a function via `cargo rustc ...` [here](https://github.com/tcdi/plrust/blob/clarify-env-var/src/user_crate/state_provisioned.rs#L39)? If that is the case, I might be missing something (which is a highly distinct possibility), but how does `PLRUST_TARGET` get "stored" in the handler such that it's passed to build method that runs `cargo rustc` externally?

Reiterating I might have the wrong mental model in my mind, lol

workingjubilee: The handler doesn't set the env var, but rather checks the env var and then generates the build target based on that env var (which currently means it passes it directly through without validation but if in the future it does validation or whatever that would also apply.

BradyBonnette: Ah ok, so it'd be up to the implementer of plrust in their own systems to ensure that it exists in their environment if they choose to specify a different target?

workingjubilee: Correct. Though whether `PLRUST_TARGET` survives the current design is somewhat up in the air.

BradyBonnette: ok, great, thanks for the clarification ❤️

BradyBonnette (approved):

PR #76, opened 2022-09-07
Include db_oid when writing/referencing the binary on disk

JohnHVancouver:
FunctionOids are unique within the database but not across it. It's possible for two PL/Rust functions to share the same fn_oid across different databases leading to overwriting the same function, or accessing it.

To prevent this include db_oid when constructing the name to guarantee uniqueness of the files

Note this might be unnecessary by the time https://github.com/tcdi/plrust/issues/28 is addressed but until then it seems like it'd be better to avoid this potential case.

### DIFF
--- a/src/plrust.rs
+++ b/src/plrust.rs
@@ -92,6 +92,8 @@ pub(crate) fn compile_function(fn_oid: pg_sys::Oid) -> eyre::Result<(PathBuf, Ou
     let work_dir = gucs::work_dir();
     let pg_config = gucs::pg_config();
     let target_dir = work_dir.join("target");
+    // SAFETY: Postgres globally sets this to `const InvalidOid`, so is always read-safe,
+    // then writes it only during initialization, so we should not be racing anyone.
     let db_oid = unsafe { MyDatabaseId };
 
     let generated = unsafe { UserCrate::try_from_fn_oid(db_oid, fn_oid)? };
--- a/src/plrust.rs
+++ b/src/plrust.rs
@@ -11,7 +11,8 @@ use crate::{
     gucs,
     user_crate::{StateLoaded, UserCrate},
 };
-use pgx::{pg_sys::FunctionCallInfo, *};
+
+use pgx::{pg_sys::FunctionCallInfo, pg_sys::MyDatabaseId, *};
 use std::{
     cell::RefCell,
     collections::{hash_map::Entry, HashMap},
@@ -47,12 +48,13 @@ pub(crate) unsafe fn evaluate_function(
 ) -> eyre::Result<pg_sys::Datum> {
     LOADED_SYMBOLS.with(|loaded_symbols| {
         let mut loaded_symbols_handle = loaded_symbols.borrow_mut();
+        let db_oid = MyDatabaseId;
         let user_crate_loaded = match loaded_symbols_handle.entry(fn_oid) {
             entry @ Entry::Occupied(_) => {
                 entry.or_insert_with(|| unreachable!("Occupied entry was vacant"))
             }
             entry @ Entry::Vacant(_) => {
-                let crate_name = crate_name(fn_oid);
+                let crate_name = crate_name(db_oid, fn_oid);
                 let mut shared_object_name = crate_name;
                 #[cfg(any(
                     all(target_os = "macos", target_arch = "x86_64"),
@@ -68,7 +70,7 @@ pub(crate) unsafe fn evaluate_function(
                 shared_object_name.push_str(DLL_SUFFIX);
 
                 let shared_library = gucs::work_dir().join(&shared_object_name);
-                let user_crate_built = UserCrate::built(fn_oid, shared_library);
+                let user_crate_built = UserCrate::built(db_oid, fn_oid, shared_library);
                 let user_crate_loaded = user_crate_built.load()?;
 
                 entry.or_insert(user_crate_loaded)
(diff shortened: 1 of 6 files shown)


src/user_crate/state_generated.rs
```diff
@@ -419,7 +439,7 @@ mod tests {
 
             let generated_lib_rs = generated.lib_rs()?;
             let fixture_lib_rs = parse_quote! {
-                use ::pgx::*;
+                use pgx::*;
```
workingjubilee, line 442: This should remain the same.

src/plrust.rs
```diff
@@ -90,8 +92,9 @@ pub(crate) fn compile_function(fn_oid: pg_sys::Oid) -> eyre::Result<(PathBuf, Ou
     let work_dir = gucs::work_dir();
     let pg_config = gucs::pg_config();
     let target_dir = work_dir.join("target");
+    let db_oid = unsafe { MyDatabaseId };
```
workingjubilee, line 95: When exactly does this global variable get initialized by Postgres? And does it conceivably get altered during runtime, or is it strictly write-once from our perspective? It would be ideal if we documented the safety conditions for this, as bindgen says this is a `static mut` and thus racy.

src/user_crate/state_generated.rs
```diff
@@ -520,7 +544,7 @@ mod tests {
...
```
workingjubilee, line 547: This should remain the same.

workingjubilee (commented):
This looks good, but I have a question about the safety conditions implied here.

src/plrust.rs
JohnHVancouver: It really only gets set once, unless some extensions do weird things but we shouldn't worry about those.
I ran against the latest postgres code:

```
desktop  % grep -ri "MyDatabaseId = " *
src/backend/utils/init/globals.c:Oid			MyDatabaseId = InvalidOid;
src/backend/utils/init/postinit.c:		MyDatabaseId = Template1DbOid;
src/backend/utils/init/postinit.c:		MyDatabaseId = dbform->oid;
src/backend/utils/init/postinit.c:		MyDatabaseId = dbform->oid;
```
the postinit.c ones are only called when a Postgres backend is initialized:

https://github.com/postgres/postgres/blob/master/src/backend/utils/init/postinit.c#L664

globals.c just initializes the value for the first time:
https://github.com/postgres/postgres/blob/master/src/backend/utils/init/globals.c#L89

src/user_crate/state_generated.rs
JohnHVancouver: ooc what's the difference? rustfmt seems to correct it so I wasn't too sure

workingjubilee: Running `cargo fmt` has never corrected this for me before.
`use ::name::{import, imports};` in practice is rarely needed. However, it is hypothetically possible for a name to be shadowed. `use ::` starts from the "top level" always, so it means that you import from pgx, the crate, and not from, say, a user module defined with the name pgx.

https://doc.rust-lang.org/reference/items/use-declarations.html#use-paths

src/plrust.rs
workingjubilee: Ah. The global definition means it is always safe to read in Rust terms (i.e. is not uninitialized data). And yes, I think it is safe to assume we're not racing in practice.
```suggestion
    // SAFETY: Postgres globally sets this to `const InvalidOid`, so is always read-safe,
    // then writes it only during initialization, so we should not be racing anyone.
    let db_oid = unsafe { MyDatabaseId };
```

src/user_crate/state_generated.rs
JohnHVancouver: Interesting, cargo fmt and rustfmt seems to work differently?
Tried cargo fmt and got what you mentioned but with rustfmt

```
dev-dsk-hsuchen-2a-9713b625 % git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean

(22-09-09 17:23:20) <0> [~/proj/plrust/src]  
dev-dsk-hsuchen-2a-9713b625 % find . -name "*.rs" -exec rustfmt {} \;

(22-09-09 17:23:24) <0> [~/proj/plrust/src]  
dev-dsk-hsuchen-2a-9713b625 % git diff
diff --git a/src/tests.rs b/src/tests.rs
index d4c213f..4fa4610 100644
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -346,7 +346,7 @@ mod tests {
     #[pg_test]
     #[search_path(@extschema@)]
     fn plrust_one_sleepy_boi() {
-        use std::{time::Duration, thread::sleep};
+        use std::{thread::sleep, time::Duration};
         let moment = Duration::from_secs(2);
         sleep(moment);
 
diff --git a/src/user_crate/state_generated.rs b/src/user_crate/state_generated.rs
index 6e1b7bc..a438085 100644
--- a/src/user_crate/state_generated.rs
+++ b/src/user_crate/state_generated.rs
@@ -127,7 +127,7 @@ impl StateGenerated {
     #[tracing::instrument(level = "debug", skip_all, fields(fn_oid = %self.fn_oid))]
         let mut skeleton: syn::File = syn::parse_quote!(
-            use ::pgx::*;
+            use pgx::*;
         );
 
         let crate_name = self.crate_name();
```

workingjubilee: They do not work differently. Rather, `cargo fmt` invokes `rustfmt` with certain flags, which it picks out from reading the Cargo.toml file. The same is also the case with all cargo commands: cargo is rustc's most adept "power user".

JohnHVancouver: Ahh thanks for the clarification. Let me make the change

workingjubilee:
Excellent, thank you!

workingjubilee (approved):
Excellent, thank you!

PR #77, opened 2022-09-07
ARM64 build testing

BradyBonnette:


workingjubilee:
Hmm. If this is functional and the blocker is optimization, shouldn't this be merged and a followup PR be opened?

BradyBonnette:
> Hmm. If this is functional and the blocker is optimization, shouldn't this be merged and a followup PR be opened?

If it is preferable that way I can absolutely do that.

workingjubilee (approved):
Excellent, it looks like everything is working! I didn't mean to interrupt your work on optimization much, I just wanted to be sure that we knew the pipeline was clear before we started trying to make it faster.

BradyBonnette:
@workingjubilee no you're right. It's probably way more imperative to get it out there and useful before I make it better.

PR #82, opened 2022-09-15
Update to use rustc 1.61

workingjubilee:


BradyBonnette (approved):
Looks like it's using 1.61 where it should be!

Issue #84, opened 2022-09-20
Decrease arm64 build times


BradyBonnette:
The custom Github Actions runners for arm64 (at the time of this writing) take about 30 minutes to build, which seems a bit excessive. In the past, there was some work done for other various projects (such as PGX) to decrease their build times by implementing various caching mechanisms. The same thing should be done here, but albeit in a different manner considering the arm64 runners are hosted in a completely different environment than the out-of-the-box Github Actions runners.

BradyBonnette:
Closing this as a result of #85 getting merged in.

PR #85, opened 2022-09-20
Speeds up arm64 build times

BradyBonnette:
All work here is an attempt to bring down the build times for plrust on arm64. Right now the builds are roughly 30 minutes give or take. I am fairly certain we can get that down to a lower number.

Addresses https://github.com/tcdi/plrust/issues/84

### DIFF
--- a/.github/scripts/load_cache.sh
+++ b/.github/scripts/load_cache.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# Downloads and extracts cache archive from S3.
+#
+# Expects the following environment variables to already exist:
+#  * ARTIFACT_USER_AWS_PROFILE: the profile to use when issuing AWS CLI commands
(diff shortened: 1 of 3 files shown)


BradyBonnette:
@workingjubilee 

As I was working on this, I noticed that the pg14 x86_64 tests run both `postgrestd` and `non-postgrestd` tests, whereas the pg13 x86_64 version only runs the `non-postgrestd` tests.

Do you think that the arm64 (pg14) builds should do both, similar to the pg14 x86_64 testing? Or should it continue to use the `postgrestd` version only? If you feel like this should change, Ill do so in another pull request once this gets merged in.

workingjubilee (approved):
Excellent work!

To answer your question: No. Now that we have both aarch64 and x86_64 building, and a slightly more settled postgrestd implementation, it's actually about time to disable building user functions with the "normal" host targets, as they will be _abnormal_ for our case.

BradyBonnette:
Understood. Thanks!

PR #87, opened 2022-09-21
Update for pgx#615: New SQL type mapping

workingjubilee:
This brings PL/Rust up to date with the latest diffs in pgx as a direct result of tcdi/pgx#615's new Rust-to-SQL-and-back type mapping, specifically lining us up with https://github.com/tcdi/pgx/commit/3894728ab8ed732a3087a72363f32ebe0ddd7c8b. Very shortly will come pgx 0.5.0-beta.1 which should settle the major diffs insofar as PL/Rust is concerned.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -973,7 +973,7 @@ dependencies = [
 [[package]]
 name = "pgx"
 version = "0.5.0-beta.0"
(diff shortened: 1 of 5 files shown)


BradyBonnette (approved):
Looks good! 👍

PR #88, opened 2022-09-21
Include architecture in name of plrust compiled objects

JohnHVancouver:
There's no guarantee that when the database is restored it'll land on the same host architecture as when the object was compiled.

This lays the ground work to at least avoid running the object in question

We replace '-' in the host triple with '_' because it's not recognized as a valid identifier

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ force_enable_x86_64_darwin_generations = []
 target_postgrestd = []
 
 [dependencies]
+current_platform = "0.2.0"
 cfg-if = "1"
 pgx = { version = "0.5.0-beta.0" }
 pgx-pg-config = { version = "0.5.0-beta.0" }
--- a/src/plrust.rs
+++ b/src/plrust.rs
@@ -21,6 +21,8 @@ use std::{
     process::Output,
 };
 
+use current_platform::CURRENT_PLATFORM;
+
 thread_local! {
     pub(crate) static LOADED_SYMBOLS: RefCell<HashMap<pg_sys::Oid, UserCrate<StateLoaded>>> = Default::default();
 }
@@ -106,7 +108,15 @@ pub(crate) fn compile_function(fn_oid: pg_sys::Oid) -> eyre::Result<(PathBuf, Ou
 }
 
 pub(crate) fn crate_name(db_oid: pg_sys::Oid, fn_oid: pg_sys::Oid) -> String {
-    let crate_name = format!("plrust_fn_oid_{}_{}", db_oid, fn_oid);
+    // Include current_platform in the name
+    // There's no guarantee that the compiled library will be
+    // in the same architecture if the database was restored
+    let crate_name = format!(
+        "plrust_fn_oid_{}_{}_{}",
+        db_oid,
+        fn_oid,
+        CURRENT_PLATFORM.replace("-", "_")
+    );
 
     crate_name
 }


JohnHVancouver:
Expanding on the `CURRENT_PLATFORM.replace("-", "_")` 

This is the error you'd get otherwise:

```
ERROR:  "plrust_fn_oid_16391_147471_x86_64-unknown-linux-gnu" is not a valid Ident
CONTEXT:  /home/hsuchen/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.39/src/fallback.rs:700:9
```

https://github.com/dtolnay/proc-macro2/blob/master/src/fallback.rs#L729-L730

I think it's erroring here: https://github.com/dtolnay/proc-macro2/blob/master/src/fallback.rs#L702-L704

where I think essentially `-` is not a valid but `_` is.

workingjubilee:
These identifiers are getting very long, I am starting to get concerned about Postgres trimming them somehow.

JohnHVancouver:
Yea it's a bit wonky but all of this might go away if https://github.com/tcdi/plrust/issues/28 is done somehow but I'd like to include this in case that doesn't get done. We could move to a per OS/DB_OID/FN_OID directory structure but I don't feel particularly strongly about it one way or another.

Postgres trims identifiers depending on the system compilation setting [0], default for identifiers is 63 bytes. IIUC identifiers refer to things like column, function, table names etc. However I don't think this counts as an identifier, and even if it is we're not storing any of these identifiers in the database anywhere. We're relying on dbOid, fnOid, and host triple to construct the name. None of which exceeds the identifier by itself

[0] https://www.postgresql.org/docs/current/limits.html

Cargo.toml
```diff
@@ -26,6 +26,7 @@ force_enable_x86_64_darwin_generations = []
 target_postgrestd = []
 
 [dependencies]
+current_platform = "0.2.0"
```
workingjubilee, line 29: This dependency holds less actual code than the `leftpad` of npm fame did.
```suggestion
```

src/plrust.rs
```diff
@@ -21,6 +21,8 @@ use std::{
     process::Output,
 };
 
+use current_platform::CURRENT_PLATFORM;
+
```
workingjubilee, line 25:
```suggestion
const CURRENT_PLATFORM: &str = env!("TARGET_PLATFORM");
```

workingjubilee (changes requested):

workingjubilee:
That's fair.

workingjubilee:
Ah, scratch that, I just noticed the `build.rs`, my review is incorrect.

workingjubilee (approved):

Cargo.toml
JohnHVancouver: hahaha that's fair

src/plrust.rs
JohnHVancouver: nit: Do you want TARGET_PLATFORM or HOST_PLATFORM?

There's no guarantee that HOST == TARGET always (even though it should be...)

workingjubilee: Eh, it doesn't matter 'cause the build.rs is annoying to also include, so it actually does makes sense to depend on another crate, thus why I merged it despite my initial grump.

And we do want the target platform, because that way we don't block a user fn from being compiled on one host and deployed on a different target.

JohnHVancouver: Ahh that's a good point. I was thinking about it mainly from the perspective of:

https://github.com/tcdi/plrust/blob/main/src/plrust.rs#L62

which you would want HOST_PLATFORM for?

PR #89, opened 2022-09-22
Fixes broken tests due to introduction of pgx::prelude

BradyBonnette:
It was noticed that a few plrust tests were broken after PGX introduced `pgx::prelude`. See here https://github.com/tcdi/pgx/pull/699

The work here aims to fix those tests.


workingjubilee (approved):
Nice!

PR #95, opened 2022-09-27
Add prototype unsafe checks in PL/Rust

workingjubilee:
This adds a new feature, `"verify"`, to PL/Rust, which is set by default at compilation time. It also makes all code written for PL/Rust more strict, opting in to RFC 2585's `unsafe_op_in_unsafe_fn` lint by default, to make it harder to escape `unsafe` detection with or without the `"verify"` feature. This is possible due to tcdi/pgx#707.

Together these are expected to take care of most ways of sneaking `unsafe` into a PL/Rust function, however this basic prototype does not handle a number of possible routes for recursion, crate dependencies, et cetera.

### DIFF
--- a/src/user_crate/mod.rs
+++ b/src/user_crate/mod.rs
@@ -12,6 +12,7 @@ pub(crate) use state_loaded::StateLoaded;
 pub(crate) use state_provisioned::StateProvisioned;
 
 use crate::PlRustError;
+#[cfg(feature = "verify")]
 use geiger;
 use pgx::{pg_sys, PgBuiltInOids, PgOid};
 use proc_macro2::TokenStream;
@@ -196,11 +197,14 @@ fn parse_source_and_deps(code_and_deps: &str) -> eyre::Result<(syn::Block, toml:
     }
 
     code_block.push_str(" }");
-
-    let fake_fn = format!("fn plrust_fn() {code_block}");
-    let unsafe_metrics = geiger::find_unsafe_in_string(&fake_fn, geiger::IncludeTests::No).unwrap();
-    if unsafe_metrics.counters.has_unsafe() {
-        panic!("detected unsafe code in this code block:\n{}", code_block);
+    #[cfg(feature = "verify")]
+    {
+        let fake_fn = format!("fn plrust_fn() {code_block}");
+        let unsafe_metrics =
+            geiger::find_unsafe_in_string(&fake_fn, geiger::IncludeTests::No).unwrap();
+        if unsafe_metrics.counters.has_unsafe() {
+            panic!("detected unsafe code in this code block:\n{}", code_block);
+        }
     }
 
     let user_dependencies: toml::value::Table =
--- a/src/user_crate/mod.rs
+++ b/src/user_crate/mod.rs
@@ -12,6 +12,7 @@ pub(crate) use state_loaded::StateLoaded;
 pub(crate) use state_provisioned::StateProvisioned;
 
 use crate::PlRustError;
+use geiger;
 use pgx::{pg_sys, PgBuiltInOids, PgOid};
 use proc_macro2::TokenStream;
 use quote::quote;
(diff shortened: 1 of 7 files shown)


src/user_crate/state_generated.rs
```diff
@@ -137,6 +137,7 @@ impl StateGenerated {
     #[tracing::instrument(level = "debug", skip_all, fields(db_oid = %self.db_oid, fn_oid = %self.fn_oid))]
     pub(crate) fn lib_rs(&self) -> eyre::Result<syn::File> {
         let mut skeleton: syn::File = syn::parse_quote!(
+            #![deny(unsafe_op_in_unsafe_fn)]
```
workingjubilee, line 140: This is what required tcdi/pgx#707.

src/user_crate/mod.rs
```diff
@@ -12,6 +12,8 @@ pub(crate) use state_loaded::StateLoaded;
 pub(crate) use state_provisioned::StateProvisioned;
 
 use crate::PlRustError;
+#[cfg(feature = "verify")]
+use geiger;
```
workingjubilee, line 16: I am not fully satisfied with the geiger crate's implementation, but for basic user fn checks, it seemed sufficient as a start.

Hoverbear (approved):
Nice job. :)

Hoverbear:
Do you think proc macro expansion should be done before running Geiger? https://github.com/rust-lang/rust/issues/90765

src/user_crate/mod.rs
```diff
@@ -194,11 +196,22 @@ fn parse_source_and_deps(code_and_deps: &str) -> eyre::Result<(syn::Block, toml:
         }
     }
 
+    code_block.push_str(" }");
+    #[cfg(feature = "verify")]
```
workingjubilee, line 200: Fundamentally I question the premise that what Postgres calls a "trusted" language would not be able to ever use what Rust calls `unsafe`. However, if used to execute _untrusted user code_, obviously it's not acceptable. In fact, a fair number of things that Postgres would probably allow a trusted user language to do are things you probably would not want to let untrusted user code do.

Thus the name "verify", as in "trust but verify".

I think there should be a few sliders, in essence, on PL/Rust's safety configurations, so you can set those at build time, install the resulting extension, and declare your extension to be `plrust` or `plrustu`, depending on your safety assessment for your database installation. We probably want to initially make things work as much as possible with a maximally secured version by default and probably make it so that almost all security features default to "on".

src/user_crate/mod.rs
```diff
@@ -194,11 +196,22 @@ fn parse_source_and_deps(code_and_deps: &str) -> eyre::Result<(syn::Block, toml:
...
+    {
+        let fake_fn = format!("fn plrust_fn() {code_block}");
```
workingjubilee, line 202: Part of the reason I am not satisfied with the geiger crate's implementation: it expects you are giving it a string of items.

workingjubilee:
> Do you think proc macro expansion should be done before running Geiger? [rust-lang/rust#90765](https://github.com/rust-lang/rust/issues/90765)

Probably, but that requires nightly... Currently this PR doesn't make dependencies safe, and a proc macro has to be another crate, anyways, so I suppose the biggest threat, thinking about it, would be expansions from PGX itself.

workingjubilee:
The latest commit adds a `// TODO` test for the `plutonium` crate that is `#[ignore]`d by default, in the hopes that this TODO will eventually draw attention and be resolved.

BradyBonnette:
👀 I need to study this one a little more to understand, only because I am curious. I know you didnt add me as a reviewer, but this is interesting.

src/user_crate/mod.rs
Hoverbear: Yeah... Hmm...

src/user_crate/mod.rs
Hoverbear: I wonder if offering "capabilities" which the admin could tweak (ala https://github.com/bytecodealliance/cap-std) might be interesting...

PR #97, opened 2022-09-28
Update to pgx 0.5.0-beta.1

workingjubilee:
Everything bumps and plrust fn are now built with the "plrust" feature flag. This also means it doesn't have to use the latest git branch as a patch anymore.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -451,6 +451,29 @@ mod tests {
         "#;
         Spi::run(definition);
     }
+
+    #[pg_test]
+    #[search_path(@extschema@)]
+    #[should_panic]
+    fn plrust_pgloglevel_dont_allcaps_panic() {
+        // This test attempts to annihilate the database.
+        // It relies on the existing assumption that tests are run in the same Postgres instance,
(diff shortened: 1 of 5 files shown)


BradyBonnette (approved):
Yessssssss. I especially love not having to rely on a branch anymore.

src/tests.rs
```diff
@@ -451,6 +451,29 @@ mod tests {
...
+
+                elog(PgLogLevel::PANIC, "If other tests completed, PL/Rust did not actually destroy the entire database, \
+                                         But if you see this in the error output, something might be wrong.");
+                Some("lol".into())
+            $$ LANGUAGE plrust;
+        "#;
+        Spi::run(definition);
+        let retval = Spi::get_one::<String>("SELECT dont_allcaps_panic();\n");
+        assert_eq!(retval, Some("lol".into()));
+    }
```
BradyBonnette, line 476: 👀

workingjubilee: I realized the code that made this "safe"-ish was added to pgx but not tested here, in effect.

PR #100, opened 2022-10-12
Updates to use pgx 0.5.3

BradyBonnette:


Cargo.toml
```diff
@@ -29,8 +29,8 @@ verify = ["dep:geiger"]
 [dependencies]
 current_platform = "0.2.0"
 cfg-if = "1"
-pgx = { version = "0.5.0-beta.1" }
```
thomcc, line 32: Nit: `~` doesn't make a difference here from the default (and anything aside from the default hits the "cargo's resolver behaves badly" behavior)

thomcc: Oh, unless this is a hack to make patch.crates-io work? I don't *think* it should matter for that but... maybe?

workingjubilee: I was mostly experimenting, tbh.

workingjubilee:
Ah yep that'll do it. Okay, just need to get to 0.5.3 then.

workingjubilee (approved):
Feel free to merge at your leisure, @BradyBonnette.

PR #102, opened 2022-10-17
Update to pgx 0.5.2

JohnHVancouver:


JohnHVancouver:
Hm is there something about the way the aarch64 tests are configured that causes it to break?

JohnHVancouver:
Closing in favour of https://github.com/tcdi/plrust/pull/100

BradyBonnette:
> Hm is there something about the way the aarch64 tests are configured that causes it to break?

My limited understanding at the moment is that it's less about the CPU architecture (aarch64) and more about the platform OS (CentOS-based is the current standing theory). I believe @thomcc did a local-to-him test using aarch64 in a (I think?) Debian-based OS and it was working fine.

The test setup between pgx pre-0.5.0 and 0.5.0 didn't change all that much save for some additional output. The breakage was in PGX 0.5.0 and it was bindgen related, see: https://github.com/tcdi/pgx/commit/d5fb78fa3674e81654bdf5a5c584f214d9b71916

PR #103, opened 2022-10-19
Provide allow_list of dependencies

JohnHVancouver:
Provide a GUC that will specify the file path
that contains an allow-listed set of dependencies.
If it is not set, treat it as allowing
an arbitrary list of dependencies. This is useful
to allow administrators of PL/Rust to limit the
scope of dependencies that are allowed in their
environment.

This has a random commit to support pgx 0.5.2 but I've removed that PR. in favour of waiting for #100 to be resolved

Posting this to get some initial feedback on the approach. There's some issues with the matching, namely that it doesn't support toml::Value::Array nicely, and it only does a strict equality check. 

For instance if the file had:

```
foo = "1.1.5"
```

and the user specified dependency as `foo = "1.1"` this would break.
I'm not sure if there's a nice crate that would allow these comparison checks to be done. But I'm hoping this is extensible enough since the validation check could be built upon.

### DIFF
--- a/build
+++ b/build
@@ -21,7 +21,6 @@ fi
 
 (
     if cd postgrestd; then
-        git pull
         git submodule update --init --recursive
     else
         git clone https://github.com/tcdi/postgrestd.git --branch "1.61" --recurse-submodules
--- a/build
+++ b/build
@@ -14,13 +14,13 @@ if [ "$CI" != true ]; then
     --version "0.6.0-alpha.1" \
     --locked # use the Cargo.lock in the pgx repo
 fi
+
 # Don't need to run cargo pgx init: user might already have set it up,
 # and doing so risks clobbering their configuration.
 # If they get an error, it's fairly self-explanatory.
 
 (
     if cd postgrestd; then
-        git pull
         git submodule update --init --recursive
     else
         git clone https://github.com/tcdi/postgrestd.git --branch "1.61" --recurse-submodules
--- a/build
+++ b/build
@@ -21,6 +21,7 @@ fi
 
 (
     if cd postgrestd; then
+        git pull
         git submodule update --init --recursive
     else
         git clone https://github.com/tcdi/postgrestd.git --branch "1.61" --recurse-submodules
--- a/build
+++ b/build
@@ -11,7 +11,7 @@ cargo update -p pgx
 cargo fetch
 if [ "$CI" != true ]; then
     cargo install cargo-pgx \
-    --git "https://github.com/tcdi/pgx.git" --branch "develop" \
+    --version "0.6.0-alpha.1" \
     --locked # use the Cargo.lock in the pgx repo
 fi
 # Don't need to run cargo pgx init: user might already have set it up,
@@ -26,6 +26,7 @@ fi
         git clone https://github.com/tcdi/postgrestd.git --branch "1.61" --recurse-submodules
         cd ./postgrestd
     fi
+    rm rust-toolchain.toml
     ./run clean
     STD_TARGET="$PLRUST_TARGET" ./run install
 )
--- a/src/gucs.rs
+++ b/src/gucs.rs
@@ -91,3 +91,29 @@ pub(crate) fn tracing_level() -> tracing::Level {
         .map(|v| v.parse().expect("plrust.tracing_level was invalid"))
         .unwrap_or(tracing::Level::INFO)
 }
+
+pub(crate) fn allow_listed_dependencies_only() -> bool {
+    let val = PLRUST_ALLOWED_DEPENDENCIES.get();
+
+    if val.is_none() {
+        return false;
+    }
+
+    true
+}
+
+pub(crate) fn get_allow_listed_dependencies() -> toml::value::Table {
+    let path = PathBuf::from_str(
+        &PLRUST_ALLOWED_DEPENDENCIES
+            .get()
+            .expect("plrust.allowed_dependencies is not set in postgresql.conf"),
+    )
+    .expect("plrust.allowed_dependencies is not a valid path");
+
+    let contents = std::fs::read_to_string(&path).expect(
+        "
+        Unable to read allow listed dependencies",
+    );
+
+    toml::from_str(&contents).expect("Unable to format allow listed dependencies")
+}
--- a/src/gucs.rs
+++ b/src/gucs.rs
@@ -7,6 +7,7 @@ All rights reserved.
 Use of this source code is governed by the PostgreSQL license that can be found in the LICENSE.md file.
 */
 
+use once_cell::sync::Lazy;
 use pgx::guc::{GucContext, GucRegistry, GucSetting};
 use std::path::PathBuf;
 use std::str::FromStr;
@@ -15,6 +16,20 @@ static PLRUST_WORK_DIR: GucSetting<Option<&'static str>> = GucSetting::new(None)
 static PLRUST_PG_CONFIG: GucSetting<Option<&'static str>> = GucSetting::new(None);
 static PLRUST_TRACING_LEVEL: GucSetting<Option<&'static str>> = GucSetting::new(None);
 static PLRUST_ALLOWED_DEPENDENCIES: GucSetting<Option<&'static str>> = GucSetting::new(None);
+
+static PLRUST_ALLOWED_DEPENDENCIES_CONTENTS: Lazy<toml::value::Table> = Lazy::new(|| {
+    let path = PathBuf::from_str(
+        &PLRUST_ALLOWED_DEPENDENCIES
+            .get()
+            .expect("plrust.allowed_dependencies is not set in postgresql.conf"),
+    )
+    .expect("plrust.allowed_dependencies is not a valid path");
+
+    let contents =
+        std::fs::read_to_string(&path).expect("Unable to read allow listed dependencies");
+
+    toml::from_str(&contents).expect("Unable to format allow listed dependencies")
+});
 
 pub(crate) fn init() {
     GucRegistry::define_string_guc(
@@ -85,18 +100,6 @@ pub(crate) fn allow_listed_dependencies_only() -> bool {
     true
 }
 
-pub(crate) fn get_allow_listed_dependencies() -> toml::value::Table {
-    let path = PathBuf::from_str(
-        &PLRUST_ALLOWED_DEPENDENCIES
-            .get()
-            .expect("plrust.allowed_dependencies is not set in postgresql.conf"),
-    )
-    .expect("plrust.allowed_dependencies is not a valid path");
-
-    let contents = std::fs::read_to_string(&path).expect(
-        "
-        Unable to read allow listed dependencies",
-    );
-
-    toml::from_str(&contents).expect("Unable to format allow listed dependencies")
+pub(crate) fn get_allow_listed_dependencies() -> &'static toml::value::Table {
+    &*PLRUST_ALLOWED_DEPENDENCIES_CONTENTS
 }
--- a/src/gucs.rs
+++ b/src/gucs.rs
@@ -14,6 +14,7 @@ use std::str::FromStr;
 static PLRUST_WORK_DIR: GucSetting<Option<&'static str>> = GucSetting::new(None);
 static PLRUST_PG_CONFIG: GucSetting<Option<&'static str>> = GucSetting::new(None);
 static PLRUST_TRACING_LEVEL: GucSetting<Option<&'static str>> = GucSetting::new(None);
+static PLRUST_ALLOWED_DEPENDENCIES: GucSetting<Option<&'static str>> = GucSetting::new(None);
 
 pub(crate) fn init() {
     GucRegistry::define_string_guc(
@@ -37,6 +38,14 @@ pub(crate) fn init() {
         "The tracing level to use while running pl/rust",
         "The tracing level to use while running pl/rust. Should be `error`, `warn`, `info`, `debug`, or `trace`",
         &PLRUST_TRACING_LEVEL,
+        GucContext::Sighup,
+    );
+
+    GucRegistry::define_string_guc(
+        "plrust.allowed_dependencies",
+        "The full path of a file containing crates and versions allowed when creating PL/Rust functions.",
+        "The full path of a file containing crates and versions allowed when creating PL/Rust functions.",
+        &PLRUST_ALLOWED_DEPENDENCIES,
         GucContext::Sighup,
     );
 }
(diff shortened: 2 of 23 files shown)


workingjubilee:
Removed the version commit.

src/user_crate/mod.rs
```diff
@@ -210,12 +211,52 @@ fn parse_source_and_deps(code_and_deps: &str) -> eyre::Result<(syn::Block, toml:
...
+                            .unwrap(),
+                    )
+                {
+                    // version is supported
+                } else {
+                    unsupported_deps.insert(dep.to_string(), version.clone());
+                }
+            }
+            _ => {
+                pgx::error!("The toml::Value {:?} is not currently supported", version);
```
workingjubilee, line 246: We should be able to handle the "inline table" format and touch the "version" key inside it, as this is commonly used by Rust programmers. You can see an example of a similar block of code here: https://github.com/tcdi/pgx/blob/99f92951f630b51e0c32f00e8db3602b43f961d0/pgx-version-updater/src/main.rs#L262-L299

This way we are prepared for e.g. when we want to qualify which features of a crate are allowed.

src/user_crate/mod.rs
```diff
@@ -210,12 +211,52 @@ fn parse_source_and_deps(code_and_deps: &str) -> eyre::Result<(syn::Block, toml:
...
+            }
+        }
+    }
+
+    if !unsupported_deps.is_empty() {
+        pgx::error!(
+            "The following dependencies are unsupported {:?}, environment only supports {:?}",
```
workingjubilee, line 253: "Environment" is vague? This can be plainer: The configured PL/Rust permissions don't allow this. That way if someone is the system administrator they know what to change (or not) to fix this.

src/gucs.rs
```diff
@@ -39,6 +55,14 @@ pub(crate) fn init() {
         &PLRUST_TRACING_LEVEL,
         GucContext::Sighup,
     );
+
+    GucRegistry::define_string_guc(
+        "plrust.allowed_dependencies",
+        "The full path of a file containing crates and versions allowed when creating PL/Rust functions.",
+        "The full path of a file containing crates and versions allowed when creating PL/Rust functions.",
```
workingjubilee, line 62: In what format?

workingjubilee (changes requested):
Sorry for taking so long to getting around to reviewing this.

src/tests.rs
```diff
@@ -495,11 +549,42 @@ pub mod pg_test {
     });
     static LOG_LEVEL: &str = "plrust.tracing_level=trace";
 
+    static PLRUST_ALLOWED_DEPENDENCIES_FILE_NAME: &str = "allowed_deps.txt";
```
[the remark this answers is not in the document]
JohnHVancouver: Makes sense, will make changes to use toml more explicitly

src/user_crate/mod.rs
JohnHVancouver: So the reason I wanted to include the existing crates that are supported is that it seems like an easy way to see what's actually there, instead of having to also expose it via documentation/a SQL function

src/user_crate/mod.rs
JohnHVancouver: Added a basic table support for "version", not familiar with what values could potentially be used in an inline table, I'm guessing it can be arbitrary, in which case we probably need to add something recursive eventually

src/user_crate/mod.rs
workingjubilee: Including the contents of the list is fine, I just want the language to be plainer.

src/gucs.rs
JohnHVancouver: specified toml

src/user_crate/mod.rs
JohnHVancouver: Ahh thanks. I was wondering if something like this was possible:

```
pgx = {version = "0.6.0", features = ["a", "b", "c"], random_crate_specific_key = ["e", "f"]}
```

But looking at the manifest it doesn't seem like it.

src/gucs.rs
```diff
@@ -7,13 +7,29 @@ All rights reserved.
...
+            .get()
+            .expect("plrust.allowed_dependencies is not set in postgresql.conf"),
+    )
+    .expect("plrust.allowed_dependencies is not a valid path");
+
+    let contents =
+        std::fs::read_to_string(&path).expect("Unable to read allow listed dependencies");
+
+    toml::from_str(&contents).expect("Unable to format allow listed dependencies")
+});
```
workingjubilee, line 32: There's a problem with this: `Lazy` is init-once. As far as I remember, that breaks the `Sighup` resetting of the rest of these. This means it is possible to get a highly invalid configuration going with PL/Rust after a sighup if this code isn't run again, which requires this value to be mutated (rerunning this code).

src/user_crate/mod.rs
```diff
@@ -210,12 +211,76 @@ fn parse_source_and_deps(code_and_deps: &str) -> eyre::Result<(syn::Block, toml:
     let user_dependencies: toml::value::Table =
         toml::from_str(&deps_block).map_err(PlRustError::ParsingDependenciesBlock)?;
 
+    if crate::gucs::allow_listed_dependencies_only() {
+        validate_dependences_are_allowed(&user_dependencies);
+    }
```
workingjubilee, line 216: Per my remarks about separating parsing/validation vs. control: I would actually rather not have an entry point decide whether to run the validation. I would rather attempt to run the validation every time, as even a "no-op" validation step would cause errors if we see a `[dependencies]` table in a format we don't expect and can't handle. That raises the question of whether `allow_listed_dependencies_only` should exist at all.

src/user_crate/mod.rs
```diff
@@ -210,12 +211,76 @@ fn parse_source_and_deps(code_and_deps: &str) -> eyre::Result<(syn::Block, toml:
...
+                            unsupported_deps.insert(dep.to_string(), val.clone());
+                        }
+                    }
+                    _ => {
+                        pgx::error!("Currently only toml Inline Table and String format is supported for dependencies");
+                    }
+                }
+            }
+            _ => {
+                pgx::error!("The toml::Value {:?} is not currently supported", val);
```
workingjubilee, line 270: These kinds of bounds should not go unenforced just because the passlist is ℵ₁.

src/user_crate/mod.rs
workingjubilee: It's possible there's a key that has a fairly "wildcard" set of entries (`features` is basically like this) but afaik all the valid keys at this level are determined ahead of time.

src/gucs.rs
```diff
@@ -7,13 +7,29 @@ All rights reserved.
...
 use std::path::PathBuf;
 use std::str::FromStr;
 
 static PLRUST_WORK_DIR: GucSetting<Option<&'static str>> = GucSetting::new(None);
 static PLRUST_PG_CONFIG: GucSetting<Option<&'static str>> = GucSetting::new(None);
 static PLRUST_TRACING_LEVEL: GucSetting<Option<&'static str>> = GucSetting::new(None);
+static PLRUST_ALLOWED_DEPENDENCIES: GucSetting<Option<&'static str>> = GucSetting::new(None);
+
+static PLRUST_ALLOWED_DEPENDENCIES_CONTENTS: Lazy<toml::value::Table> = Lazy::new(|| {
+    let path = PathBuf::from_str(
```
workingjubilee, line 21: Kinda wonder if I can make the GucSettings hold `&'static Path`... I'll look into that later.

src/gucs.rs
JohnHVancouver: Yea that's a good point. I wanted to avoid reading a file/parsing it every time a function is created but that seems small in the grand scheme of things

src/user_crate/mod.rs
JohnHVancouver: I think I'm misunderstanding here.
IIUC this means that if the `toml::Value` doesn't match `String` or `Table`, we error out because we don't know how to interpret something like

```
[dependencies]
pi = 3.14
foo = true
```

src/gucs.rs
workingjubilee: https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html#method.get_mut ?

src/user_crate/mod.rs
workingjubilee: I mean control flow should not arrive here only when we plan to restrict the number of dependencies, or else submitting an unhandled form, and then restricting dependencies in a way that affects zero dependencies, causes an unexpected UX: the dependency list is now rejected.

src/user_crate/mod.rs
```diff
@@ -210,12 +211,76 @@ fn parse_source_and_deps(code_and_deps: &str) -> eyre::Result<(syn::Block, toml:
...
     let user_code: syn::Block =
         syn::parse_str(&code_block).map_err(PlRustError::ParsingCodeBlock)?;
 
     Ok((user_code, user_dependencies))
 }
 
+#[tracing::instrument(level = "debug", skip_all)]
+fn validate_dependences_are_allowed(user_dependencies: &toml::value::Table) {
+    let allowed_deps = crate::gucs::get_allow_listed_dependencies();
+    let mut unsupported_deps = std::collections::HashMap::<String, toml::value::Value>::new();
```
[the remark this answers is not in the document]
workingjubilee: My concern is entirely about `unsupported_deps` actually.

src/gucs.rs
```diff
@@ -65,3 +89,17 @@ pub(crate) fn tracing_level() -> tracing::Level {
...
+
+pub(crate) fn allow_listed_dependencies_only() -> bool {
+    let val = PLRUST_ALLOWED_DEPENDENCIES.get();
+
+    if val.is_none() {
+        return false;
+    }
+
+    true
+}
```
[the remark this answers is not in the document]
JohnHVancouver: Well, it seems like it's still useful to check if we should read the dependencies file at all?

src/user_crate/mod.rs
```diff
@@ -210,12 +211,76 @@ fn parse_source_and_deps(code_and_deps: &str) -> eyre::Result<(syn::Block, toml:
...
+    }
+
     let user_code: syn::Block =
         syn::parse_str(&code_block).map_err(PlRustError::ParsingCodeBlock)?;
 
     Ok((user_code, user_dependencies))
 }
 
+#[tracing::instrument(level = "debug", skip_all)]
+fn validate_dependences_are_allowed(user_dependencies: &toml::value::Table) {
```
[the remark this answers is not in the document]
JohnHVancouver: Hmm I'm still trying to grok this a little bit.

I understand the part about wanting to run dependencies formatting always against the user_dependencies so we error out earlier. 

Think I'm having trouble separating the "parsing of dependencies" from "is this dependency in the allowed list"

IIUC even if I have something like

```
let user_dependencies: toml::Value::Table = validate_dependencies(user_deps)?;
let allowed_dependencies: toml::Value::Table = validate_dependencies(allowed_deps)?;

check_dependencies_are_allowed(user_dependencies, allowed_dependencies)?;
```

Even if `validate_dependencies` already does the parsing and checks that it only contains say `toml::Value::String` and `toml::Value::Table`

and the table is formatted correctly (e.g. only contains version and optional feature)

In `check_dependencies_are_allowed` wouldn't I still need to check/ "re-parse" for unsupported toml::Value due to how match works?

src/gucs.rs
JohnHVancouver: Thanks! Thinking about this more, there seems to be two? cases where the file contents can be stale:

1. SIGHUP points the config to a different file
2. The contents of the file has been updated, but the file remains the same

For 1) you could just cache the previous filename and check if the GUC has changed to decide when to re-read the file, but for 2) if the file contents has been updated the only way to tackle that would be to re-read it every time? In which case maybe we should just read it every time if it's a case we want to handle.

workingjubilee:
Oh psh. That's why, it wants a `&[u8]`, silly me. That's answered with `.as_bytes()`

src/gucs.rs
```diff
@@ -7,13 +7,31 @@ All rights reserved.
...
 use std::path::PathBuf;
 use std::str::FromStr;
 
 static PLRUST_WORK_DIR: GucSetting<Option<&'static str>> = GucSetting::new(None);
 static PLRUST_PG_CONFIG: GucSetting<Option<&'static str>> = GucSetting::new(None);
 static PLRUST_TRACING_LEVEL: GucSetting<Option<&'static str>> = GucSetting::new(None);
+pub(crate) static PLRUST_ALLOWED_DEPENDENCIES: GucSetting<Option<&'static str>> =
+    GucSetting::new(None);
+
+pub(crate) static PLRUST_ALLOWED_DEPENDENCIES_CONTENTS: Lazy<toml::value::Table> =
```
JohnHVancouver, line 21: I still didn't address the Lazy portion here with SIGHUP cause I wasn't sure if we wanted to handle the case where the file is the same, but the contents have changed or not?

src/gucs.rs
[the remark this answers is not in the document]
workingjubilee: That part is fair I suppose.

src/gucs.rs
workingjubilee: There are almost 100,000 crates on crates.io, so... I am not against reading it every time in principle, but I am not sure if it is actually a good idea to simply blindly open it each time.

src/user_crate/mod.rs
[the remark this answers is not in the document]
workingjubilee: It should probably be more like a top-level function `enact_dependency_policy_with_list` (or whatever other name) calling the function `check_dependencies_against_list` instead.

src/gucs.rs
workingjubilee: There are almost 100,000 crates on crates.io, so... I am not against reading it every time in principle, but I am not sure if it is actually a good idea to simply blindly open it each time.

JohnHVancouver: Actually I think the proper way to handle this would be, assuming it's possible, to open up the pgx guc api: https://github.com/tcdi/pgx/blob/master/pgx/src/guc.rs#L220

to expose GucStringCheckHook and other variants:
https://github.com/postgres/postgres/blob/f13b2088fa2d4455936e65459b77698a4452f932/src/include/utils/guc.h#L177-L181

And then we can define a function that when the hook is called, we validate the file exists, read it, and update our value. That would avoid us having to figure out when a SIGHUP is actually issued to re-read the file contents.

What do you think? As a (sad) workaround/TODO I could re-define it as PGC_POSTMASTER instead of PGC_SIGHUP since Eric's patch makes that possible now with it being loaded in shared_preload_libraries

src/user_crate/mod.rs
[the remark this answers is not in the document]
JohnHVancouver: I updated with the latest revision let me know if it makes sense/what you think is the better way

src/gucs.rs
workingjubilee: Yeah, I'm good with accepting a temporary fix for now and digging into doing it "properly" a bit later. Using the Guc hook sounds interesting/useful.

> Good point...I was thinking it'd be in the low 10s of crates that an admin would want.

Rust programmers tend to do an unusual amount of grepping and otherwise programmatically manipulating the entire source codebase of crates.io, so my inclination is to assume it's entirely plausible someone managing a PL/Rust instance will do crazy things like list every single crate minus ~100, instead.

workingjubilee: Also I should note that "do silly things" is precisely a case where I am not concerned about providing optimal performance, I just don't want to assume unwarrantedly that a number that could easily be between 0 and 100k will necessarily be only 1-50.

JohnHVancouver: Makes sense, thanks! I've updated it to PGC_Postmaster as a temporary workaround, we can come back to it when GucHooks are supported

workingjubilee (approved):

build
```diff
@@ -14,13 +14,13 @@ if [ "$CI" != true ]; then
     --version "0.6.0-alpha.1" \
     --locked # use the Cargo.lock in the pgx repo
 fi
+
 # Don't need to run cargo pgx init: user might already have set it up,
 # and doing so risks clobbering their configuration.
 # If they get an error, it's fairly self-explanatory.
 
 (
     if cd postgrestd; then
```
workingjubilee, line 23:
```suggestion
    if cd postgrestd; then
        git pull
```

workingjubilee (commented):
Oh, one last thing, apparently, just gonna fix this.

PR #104, opened 2022-10-21
Use pgx 0.5.6

workingjubilee:


workingjubilee:
I don't actually expect the diffs between 0.5.3 and 0.5.6 to be severe, either way.

BradyBonnette (approved):
Nice 👍

PR #108, opened 2022-11-03
Adding further design documentation

workingjubilee:


doc/designing_for_trust.md
```diff
@@ -0,0 +1,224 @@
...
+The first only matters if the invalid state can potentially be reached later in the program. Thus, a destructor is only relevant if the program continues.
+Otherwise, it usually takes unsound code to observe states in something that has been subject to `mem::forget`, as that function consumes the object in question.
+Everything moved into it should be "gone" and no longer visible to the rest of the program state unless `unsafe` has been used.
+The second only matters if the resources are not reclaimed by some other mechanism, such as e.g. process termination.
+
+Thus, if an error inside a function causes Rust to leave behind all of the code that contains broken state,
+then at least hypothetically not running any destructors in that code whatsoever, even if it does not include `mem::forget`, is "not a problem".
+This implies that in response to errors from Postgres, a "correct" (in the sense of "sound, not violating Rust's memory safety properties")
+error handling strategy for PL/Rust is simple abject submission that terminates the function instantly.
```
thomcc, line 133: This isn't correct. If you have an object on the stack, Drop is guaranteed to run if the process isn't exited and you don't forget it (or more broadly, move it somewhere into the control of something which may not Drop it). This is required for the pinning invariant to work, as well as for the soundness of numerous patterns in the standard library internals and the broader ecosystem.

IOW, "You cannot rely on Drop because `forget` and `abort` exist" does not mean it is sound for a Rust implementation to fail to destroy an object without running its destructor when the process lives on.

More broadly, what you're suggesting is equivalent (both in theory and in fact in implementation) to forced unwinding across functions with landing pads, and that's just definitely UB.

That said, as of the recent PGX changes, I suspect we shouldn't need to do this anymore? We probably can just use the normal panic runtime?

thomcc (commented):
This is mostly good. But yeah, the way that pl/rust implements panic is definitely broken/unsound. In practice the cases where this happens are:

- Internals of collection types (even simple ones like Vec), which put structures on the stack which clean up temporarily invalid states .

- Code that wants to guard against panics in FFI will often put a struct on the stack to ensure a panic doesn't cross extern "C", and forget it at the end.

- Anything using Pin -- pin has the invariant a Pinned object must have its destructor run unless the memory is never going to be reused (IOW, you can leak a `Pin<Box<T>>`, but that's fine -- if it's node added to a global intrusive doubly-linked-list, it just won't ever get removed, but the links will stay valid). Pin, and this use of it, is most common in async and multithreaded code, which aren't entirely relevant here.

Another problem this causes though, is that code that takes Mutex<T> or RwLock<T> (even though we allow only single-threaded use, this may be required to have safe global state, although thinking about it, this causes big problems for `palloc`...) will fail to unlock them at the end of the scope.

I also suspect the next version of PGX might allow us to fix this and use normal unwinding? But I'm not totally sure.

doc/designing_for_trust.md
workingjubilee: I have definitely learned an unfortunate amount about unwinding to the point I have concluded that I was probably wrong when I initially set out, though I still have to spend some time to wrap my head around the Pin/Unpin nonsense and thus rewrite this section. But yeah, the underlying fixes in PGX can allow us to fix this, and I still think starting from the premise of "we must override GlobalAlloc" was Good, Actually, as it allows us other advantages.

thomcc: > I still think starting from the premise of "we must override GlobalAlloc" was Good, Actually, as it allows us other advantages.

I'm not sure. What about cases where rust code stores data allocated from the `#[global_allocator] in a static and uses it across calls? This is safe if you use a Mutex/RwLock/lazy_static!/OnceCell/... to store it

Won't that be memory-unsafe as the memory will be allocated out of the current context?

workingjubilee: OH YEAH I'm expecting that we may have to do some seriously silly stuff about `static` involving context control, but that we *can*, because statics **must** be globally unique and thus we can find the damn things, one way or another.

thomcc:
> OH YEAH I'm expecting that we may have to do some seriously silly stuff about static involving context control

Seems very difficult with statics in dependencies. It's not weird to have a library do something like

```rs
use once_cell::sync::Lazy;
static SOME_CACHED_ENV_VAR: Lazy<String> = Lazy::new(|| std::env::var(...).unwrap());
```

Or one of a bunch of other things which currently will cause problems due to `palloc`. We use a bunch of libraries in `pgx` that contain global statics like this, and do similar things in `pgx` itself.

I think this document should try to better justify the choice to use `palloc`. I'm guessing it's so that we can use native postgres error handling (read: longjmping over rust frames) for `panic` without leaking, but... After https://github.com/tcdi/pgx/pull/834 that shouldn't be needed, so it feels like genuinely it might be worth reconsidering. (Not to mention, `PgTryBuilder`'s implementation requires `panic` be actual `panic`s...)

eeeebbbbrrrr:
Going to go ahead and merge this.  It looks good to me.  There's definitely some more stuff we want to tackle documenting, and perhaps a few decisions we want to re-examine, but that shouldn't stand in the way of getting this out there.

PR #112, opened 2022-11-09
Use `Vec<syn::FnArg>` for deterministic arg order 

### LINKED ISSUE #111: Parameters don't line up with the names properly
When 2 of the same data type are passed in, the names don't match the ordinal number correctly.  

```
CREATE OR REPLACE FUNCTION test(a int, b int) 
  RETURNS int AS 
$$
  Some(a.unwrap())
$$ LANGUAGE plrust;

=# select test(1, 2);
 test 
------
    2
(1 row)

```

workingjubilee:
`HashMap<K, V, S>`, like most hash-keyed collection types,
does not order entries deliberately, instead simply hashing keys.
Unlike most hashed types, `HashMap<K, V, S = RandomState>` defaults
to using SipHash with a random seed to its hashing logic, which
can result in an essentially nondeterministic order in the map.
Even more peculiarly, `HashMap<K, V, RandomState>` iterates randomly,
because it does not use the key's ordering to iterate, but
simply starts at one entry and loops "in order" (i.e. random order).

By collecting arguments into a `HashMap`, then unpacking that map,
they were given a different ordinal position. This resulted in
calling with different orderings of arguments, as the arguments
only acquire names inside the Rust `fn` body: before that they
have only a number and a calling convention in the `extern "C"` ABI.
The `HashMap` is fruitless as we don't bother with the keys, so:
simply use a `Vec` of values to assure a consistent order.

This fixes #111.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -474,6 +474,103 @@ mod tests {
         let retval = Spi::get_one::<String>("SELECT dont_allcaps_panic();\n");
         assert_eq!(retval, Some("lol".into()));
     }
+
+    #[pg_test]
+    #[search_path(@extschema@)]
+    fn plrust_call_1st() {
+        let definition = r#"
+            CREATE OR REPLACE FUNCTION ret_1st(a int, b int)
+            RETURNS int AS
+            $$
+                a
+            $$ LANGUAGE plrust;
+        "#;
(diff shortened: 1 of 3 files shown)


thomcc (approved):
LGTM

PR #114, opened 2022-11-11
Add boolean guc to enable auto-recompile upon missing so file (resolves #113)

### LINKED ISSUE #113: Recompile function automatically if function .so file is not found
In cases where the .so file is not found, such as cases where the .so file is only created on writer but not on replica, or cases where the machines somehow switches to use a CPU on another architecture (e.g. x86 -> aarch64, vice versa). We'd need to use `plrust.recompile_function` to recompile all the existing functins. 

We can introduce a boolean guc which controls whether plrust would recompile a function if it found that the .so file is missing.

anth0nyleung:
## Issue - https://github.com/tcdi/plrust/issues/113

This commit adds a boolean guc which controls whether plrust will recompile missing .so files automatically when it is not found in plrust.workdir. By default it is set to false, but it can be updated dynamically.

By default
```
postgres=# show plrust.auto_recompile;
-[ RECORD 1 ]---------+----
plrust.auto_recompile | off
```

Remove the .so file for a function in `plrust.work_dir`, calling the function will then fail
```
postgres=# select foo(1);
ERROR:
   0: /home/antholeu/bc/plrust_workdir/plrust_fn_oid_13949_16396_x86_64_unknown_linux_gnu.so: cannot open shared object file: No such file or directory

Location:
   src/user_crate/state_loaded.rs:30

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
CONTEXT:  src/lib.rs:105:21
postgres=#
```

Set plrust.auto_recompile to true
```
postgres=# show plrust.auto_recompile;
-[ RECORD 1 ]---------+---
plrust.auto_recompile | on
```

Invoke the function
```
postgres=# select foo(1);
-[ RECORD 1 ]
foo | 1
```

### DIFF
--- a/src/gucs.rs
+++ b/src/gucs.rs
@@ -14,6 +14,7 @@ use std::str::FromStr;
 static PLRUST_WORK_DIR: GucSetting<Option<&'static str>> = GucSetting::new(None);
 static PLRUST_PG_CONFIG: GucSetting<Option<&'static str>> = GucSetting::new(None);
 static PLRUST_TRACING_LEVEL: GucSetting<Option<&'static str>> = GucSetting::new(None);
+static PLRUST_AUTO_RECOMPILE: GucSetting<bool> = GucSetting::new(false);
 
 pub(crate) fn init() {
     GucRegistry::define_string_guc(
@@ -39,6 +40,14 @@ pub(crate) fn init() {
         &PLRUST_TRACING_LEVEL,
         GucContext::Sighup,
     );
+
+    GucRegistry::define_bool_guc(
+        "plrust.auto_recompile",
+        "Recompile function if unable to find shared object file in plrust.work_dir",
+        "Recompile function if unable to find shared object file in plrust.work_dir",
+        &PLRUST_AUTO_RECOMPILE,
+        GucContext::Sighup,
+    )
 }
 
 pub(crate) fn work_dir() -> PathBuf {
@@ -64,4 +73,8 @@ pub(crate) fn tracing_level() -> tracing::Level {
         .get()
         .map(|v| v.parse().expect("plrust.tracing_level was invalid"))
         .unwrap_or(tracing::Level::INFO)
+}
+
+pub(crate) fn auto_recompile() -> bool {
+    PLRUST_AUTO_RECOMPILE.get()
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,6 +93,7 @@ unsafe fn plrust_call_handler(fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum
         }
         .ok_or(PlRustError::NullFmgrInfo)?
         .fn_oid;
+        unsafe { plrust::recompile_missing_shared_object(fn_oid)?; }
         let retval = unsafe { plrust::evaluate_function(fn_oid, fcinfo)? };
         Ok(retval)
     }
--- a/src/plrust.rs
+++ b/src/plrust.rs
@@ -44,6 +44,23 @@ pub(crate) unsafe fn unload_function(fn_oid: pg_sys::Oid) {
 }
 
 #[tracing::instrument(level = "debug")]
+pub(crate) unsafe fn recompile_missing_shared_object(fn_oid: pg_sys::Oid) -> eyre::Result<()> {
+    let db_oid = unsafe { MyDatabaseId };
+    let mut shared_object_name = crate_name(db_oid, fn_oid);
+    shared_object_name.push_str(DLL_SUFFIX);
+    let shared_library = gucs::work_dir().join(&shared_object_name);
+    if !shared_library.exists() && gucs::auto_recompile() {
+        tracing::trace!(
+            "Shared object {shared_object} is not found and plrust.auto_recompile is enabled, recompiling {shared_object}",
+            shared_object = shared_object_name
+        );
+        unsafe { unload_function(fn_oid); }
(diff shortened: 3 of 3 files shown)


workingjubilee:
Is there a case where we wouldn't want to simply recompile every time something is missing?

anth0nyleung:
@eeeebbbbrrrr has mentioned that there could be concurrency concerns with compiling outside of the original CREATE FUNCTION statement, so i make this configurable instead of enabled by default.

src/plrust.rs
```diff
@@ -43,6 +43,23 @@ pub(crate) unsafe fn unload_function(fn_oid: pg_sys::Oid) {
     })
 }
 
+#[tracing::instrument(level = "debug")]
+pub(crate) unsafe fn recompile_missing_shared_object(fn_oid: pg_sys::Oid) -> eyre::Result<()> {
+    let db_oid = unsafe { MyDatabaseId };
```
workingjubilee, line 48: This repeats the function logic inside `evaluate_function`, thus manages to dodge safety comments on this.

src/plrust.rs
```diff
@@ -43,6 +43,23 @@ pub(crate) unsafe fn unload_function(fn_oid: pg_sys::Oid) {
...
+    let mut shared_object_name = crate_name(db_oid, fn_oid);
+    shared_object_name.push_str(DLL_SUFFIX);
+    let shared_library = gucs::work_dir().join(&shared_object_name);
+    if !shared_library.exists() && gucs::auto_recompile() {
+        tracing::trace!(
+            "Shared object {shared_object} is not found and plrust.auto_recompile is enabled, recompiling {shared_object}",
+            shared_object = shared_object_name
```
workingjubilee, line 55: This is going to some effort to avoid allowing the formatting macro to simply capture the function.
```suggestion
    let mut shared_object = crate_name(db_oid, fn_oid);
    shared_object.push_str(DLL_SUFFIX);
    let shared_library = gucs::work_dir().join(&shared_object);
    if !shared_library.exists() && gucs::auto_recompile() {
        tracing::trace!(
            "Shared object {shared_object} is not found and plrust.auto_recompile is enabled, recompiling {shared_object}",
```

src/plrust.rs
```diff
@@ -43,6 +43,23 @@ pub(crate) unsafe fn unload_function(fn_oid: pg_sys::Oid) {
...
+        );
+        unsafe { unload_function(fn_oid); }
```
workingjubilee, line 57: The other way around formats much worse. In addition I would like a `// SAFETY` remark here that explains why we have to unload the function.
```suggestion
        unsafe { unload_function(fn_oid) };
```

src/gucs.rs
```diff
@@ -65,3 +74,7 @@ pub(crate) fn tracing_level() -> tracing::Level {
         .map(|v| v.parse().expect("plrust.tracing_level was invalid"))
         .unwrap_or(tracing::Level::INFO)
 }
+
+pub(crate) fn auto_recompile() -> bool {
+    PLRUST_AUTO_RECOMPILE.get()
+}
```
workingjubilee, line 80: This does no computational effort and thus does not need a wrapper function like this, simply use `gucs::PLRUST_AUTO_RECOMPILE.get()` elsewhere.

src/lib.rs
```diff
@@ -93,6 +93,7 @@ unsafe fn plrust_call_handler(fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum
         }
         .ok_or(PlRustError::NullFmgrInfo)?
         .fn_oid;
+        unsafe { plrust::recompile_missing_shared_object(fn_oid)?; }
```
workingjubilee, line 96: Shouldn't this happen transparently during evaluation instead of being called in sequence? It repeats a fair amount of the logic inside `evaluate_function`'s Vacant case...? There's a few cases where that can probably be factored out.

workingjubilee (changes requested):
I'm aware of the concurrency issues but I feel doubtful that making this configurable will improve them. Nonetheless, I have some comments, and I would prefer to see a test with this PR that, rather than running "normally" against PL/Rust, operates more like the `full_workflow` test in `user_crate`: artificially contriving to create a situation where the shared object isn't available but the state is otherwise as expected.

Otherwise I feel doubtful this is working correctly and am concerned about... well, as you mentioned. Concurrency concerns.

eeeebbbbrrrr:
I don't think pl/rust wants this capability at all.  The short of it is, if the .so is missing from disk, something bad happened on the host and pl/rust isn't the right tool to make decisions about what should happen next.

In my mind, I don't view a missing .so as anything different than a random database file in $PG_DATA/base/{oid}/ going missing -- and Postgres surely isn't going to try and magically re-create that.  It's gonna raise an error of some kind/severity.

Additionally, there's significant synchronization issues around this that in order to fix properly would necessitate managing shared memory locks -- one more thing I don't believe pl/rust should be doing.

---

Over on discord it was mentioned that this might improve the story around replication and re-compiling for different platforms.  That's a replication problem and we ought to be looking into solving this with Postgres' replication system -- which I'm going to investigate today.

I have a theory that pl/rust can emit a custom WAL record during CREATE FUNCTION, and then run custom code to re-create the function when Postgres replays that record.  I don't know if this is possible but I'm going to look.

In the interim, I'm going to close this PR (without merging) as it's not the right answer.

PR #115, opened 2022-11-11
Map `NUMERIC` to pgx 0.6's `AnyNumeric`

eeeebbbbrrrr:
We can't map to a specific `Numeric<P, S>` as Postgres doesn't store the "typmod" values in the `pg_proc` catalog, so by the time we create the function, those are lost to us.

I don't view this as a big deal, tho.  Users can use `AnyNumeric::rescale<P, S>().into()` to return a `AnyNumeric` that's been truncated/expanded to a particular precision and scale.  Alternatively, it can be handled at the SQL level, which I suspect is what most people do today with languages like plpgsql.

NOTE:  This PR will require a version of pgx > 0.5.6.


workingjubilee (approved):
Let's Get This Merged~

PR #121, opened 2022-11-17
Move the ultimate storage of a plrust compiled function from disk to a table

eeeebbbbrrrr:
The compilation process now cleans up after itself (we no longer leave any function-specific artifacts on disk, other than the shared `${PLRUST.WORK_DIR}target/` directory), and the .so bytes are inserted into the new `plrust.plrust_proc` table, tagged with the host's "arch" and "os", as the compiled pl/rust extension understands them.

There's more work to do around CREATE **OR REPLACE** FUNCTION, ALTER FUNCTION, and DROP FUNCTION, and likely some decisions to be changed, but putting this up for review and merge as it's otherwise compatible in functionality with what we have now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -427,6 +427,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4443176a9f2c162692bd3d352d745ef9413eec5782a80d8fd6f8a1ac692a07f7"
 
 [[package]]
+name = "fastrand"
+version = "1.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a7a407cfaa3385c4ae6b23e84623d48c2798d06e3e6a1878f7f59f17b3f86499"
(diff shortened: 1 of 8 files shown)


eeeebbbbrrrr:
Maybe this isn't quite as functionally equivalent as I thought -- the test(s) that test CREATE **OR REPLACE** fail now.  That's expected, but it's definitely a regression.  I'm going to address that and the other DDL issues in separate PRs based off this one.

eeeebbbbrrrr:
scratch that, the CREATE **OR REPLACE** bit is now in this PR (https://github.com/tcdi/plrust/pull/121/commits/7449fa49864351288d89c4fa2ecb4c8c7c0d8567).  It was just a matter of changing the SQL to UPDATE the existing row on a pkey violation.

eeeebbbbrrrr:
Yanking this in favor of a PR I haven't created yet...

PR #122, opened 2022-11-18
Store compiled shared object files in a new extension table

eeeebbbbrrrr:
Move the ultimate storage of a plrust compiled function from disk to an extension table named `plrust.plrust_proc`.

The compilation process now cleans up after itself (we no longer leave any function-specific artifacts on disk, other than the shared `${PLRUST.WORK_DIR}target/` directory), and the .so bytes are inserted into the new `plrust.plrust_proc` table, tagged with the host's "target_triple", as the compiled pl/rust extension understands it.

Also enables plrust to properly work with `DROP FUNCTION`, `DROP SCHEMA`, and `ALTER FUNCTION` when a `LANGUAGE plrust` function is involved in one of those statements.  For the first two, we remove the corresponding entires in our `plrust.plrust_proc` table, for the latter we let everything pass except changes to the `STRICT` property.

The interception of those utility statements is done via a hook, which now means plrust needs to be configured as (at least) a `shared_preload_libraries` entry in `postgresql.conf`.

`CREATE OR REPLACE FUNCTION` is simply handled through the normal language handler API.

When we load a "function.so" for the first execution, it is bytes are streamed from `plrust.plrust_proc.so` to a temporary file in the configured `plrust.work_dir` directory, dlopen()'d and then immediately deleted from disk.

Loading a function necessitates that we have one already compiled for the current host "triple target".  If one doesn't exist currently plrust will raise an ERROR.  A future PR will address an automatic re-compilation scheme.

Also of note, plrust previously would compile the user function with `-Ctarget-cpu=native`.  This is no longer the case as it needs binaries to be compatible across same-target-triple-but-different-and-incompatible-cpu-capabilities.  Perhaps in the future this can be addressed in a more explicit manner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,14 +24,22 @@ sum_array
 */
 ```
 
-# Options
+# Configuration
 
-There are two `postgresql.conf` settings that must be configured:
+First, plrust must be configured as a `shared_preload_libraries` entry in `postgresql.conf`.  For example:
+
+```
+shared_preload_libraries = 'plrust'
+```
+
+Failure to do so will cause unexpected behavior around the DROP FUNCTION, DROP SCHEMA, and ALTER FUNCTION commands.
+
+Additionally, there are two `postgresql.conf` settings that must be configured:
 
 Option | Description
 --------------|-----------
-`plrust.pg_config` | The full path of the `pg_config` binary.
-`plrust.work_dir` | The directory where pl/rust will build functions with cargo.
+`plrust.pg_config` | The full path of the `pg_config` binary (required).
+`plrust.work_dir` | The directory where pl/rust will build functions with cargo (required).
 `plrust.tracing_level` | A [tracing directive][docs-rs-tracing-directive]. (Default `info`)
 
 [github-pgx]: https://github.com/zombodb/pgx
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,7 @@
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+    println!(
+        "cargo:rustc-env=TARGET={}",
+        std::env::var("TARGET").unwrap()
+    );
+}
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -19,8 +19,55 @@ pub(crate) fn init() {
     }
 }
 
+#[cfg(feature = "pg13")]
 #[pg_guard]
 unsafe extern "C" fn plrust_process_utility_hook(
+    pstmt: *mut pg_sys::PlannedStmt,
+    query_string: *const ::std::os::raw::c_char,
+    context: pg_sys::ProcessUtilityContext,
+    params: pg_sys::ParamListInfo,
+    query_env: *mut pg_sys::QueryEnvironment,
+    dest: *mut pg_sys::DestReceiver,
+    qc: *mut pg_sys::QueryCompletion,
+) {
+    plrust_process_utility_hook_internal(
+        pstmt,
+        query_string,
+        false,
+        context,
+        params,
+        query_env,
+        dest,
+        qc,
+    )
+}
+
+#[cfg(not(feature = "pg13"))]
+#[pg_guard]
+unsafe extern "C" fn plrust_process_utility_hook(
+    pstmt: *mut pg_sys::PlannedStmt,
+    query_string: *const ::std::os::raw::c_char,
+    read_only_tree: bool,
+    context: pg_sys::ProcessUtilityContext,
+    params: pg_sys::ParamListInfo,
+    query_env: *mut pg_sys::QueryEnvironment,
+    dest: *mut pg_sys::DestReceiver,
+    qc: *mut pg_sys::QueryCompletion,
+) {
+    plrust_process_utility_hook_internal(
+        pstmt,
+        query_string,
+        read_only_tree,
+        context,
+        params,
+        query_env,
+        dest,
+        qc,
+    )
+}
+
+#[allow(unused_variables)] // for `read_only_tree` under pg13
+unsafe fn plrust_process_utility_hook_internal(
     pstmt: *mut pg_sys::PlannedStmt,
     query_string: *const ::std::os::raw::c_char,
     read_only_tree: bool,
@@ -63,6 +110,7 @@ unsafe extern "C" fn plrust_process_utility_hook(
         prev_hook(
             pstmt.into_pg(),
             query_string,
+            #[cfg(not(feature = "pg13"))]
             read_only_tree,
             context,
             params,
@@ -76,6 +124,7 @@ unsafe extern "C" fn plrust_process_utility_hook(
         pg_sys::standard_ProcessUtility(
             pstmt.into_pg(),
             query_string,
+            #[cfg(not(feature = "pg13"))]
             read_only_tree,
             context,
             params,
--- a/src/hooks.rs
+++ b/src/hooks.rs
@@ -1,6 +1,6 @@
+#![deny(unsafe_op_in_unsafe_fn)]
 use std::ffi::CStr;
 
-use pgx::pg_sys::DropStmt;
 use pgx::{
     pg_guard, pg_sys, FromDatum, IntoDatum, PgBox, PgBuiltInOids, PgList, PgLogLevel,
     PgSqlErrorCode, Spi,
@@ -67,7 +67,7 @@ unsafe extern "C" fn plrust_process_utility_hook(
 }
 
 #[allow(unused_variables)] // for `read_only_tree` under pg13
-unsafe fn plrust_process_utility_hook_internal(
+fn plrust_process_utility_hook_internal(
     pstmt: *mut pg_sys::PlannedStmt,
     query_string: *const ::std::os::raw::c_char,
     read_only_tree: bool,
@@ -77,15 +77,26 @@ unsafe fn plrust_process_utility_hook_internal(
     dest: *mut pg_sys::DestReceiver,
     qc: *mut pg_sys::QueryCompletion,
 ) {
-    let pstmt = PgBox::from_pg(pstmt);
-    let utility_stmt = PgBox::from_pg(pstmt.utilityStmt);
+    let pstmt = unsafe {
+        // SAFETY:  Postgres will have provided us with a valid PlannedStmt pointer, which it allocated
+        PgBox::from_pg(pstmt)
+    };
+    let utility_stmt = unsafe {
+        // SAFETY:  and this is the "process utility hook", so pstmt's `utilityStmt` member will also
+        // be a properly-allocated Postgres pointer
+        PgBox::from_pg(pstmt.utilityStmt)
+    };
 
     // examine the UtilityStatement itself.  We're interested in three different commands,
     // "DROP FUNCTION", "DROP SCHEMA", and "ALTER FUNCTION".  Two different node types are used
     // for these -- `DropStmt` and `AlterFunctionStmt`
 
     if utility_stmt.type_ == pg_sys::NodeTag_T_DropStmt {
-        let drop_stmt = PgBox::from_pg(pstmt.utilityStmt.cast::<pg_sys::DropStmt>());
+        let drop_stmt = unsafe {
+            // SAFETY:  we already determined that pstmt.utilityStmt is valid, and we just determined
+            // its "node type" is a DropStmt, so the cast is clean
+            PgBox::from_pg(pstmt.utilityStmt.cast::<pg_sys::DropStmt>())
+        };
 
         // in the case of DropStmt, if the object being dropped is a FUNCTION or a SCHEMA, we'll
         // go off and handle those two cases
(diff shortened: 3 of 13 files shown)


src/user_crate/mod.rs
```diff
@@ -71,24 +71,27 @@ impl UserCrate<StateProvisioned> {
...
+        self.0.crate_dir()
+    }
 }
 
 impl UserCrate<StateBuilt> {
     #[tracing::instrument(level = "debug")]
-    pub(crate) fn built(db_oid: pg_sys::Oid, fn_oid: pg_sys::Oid, shared_object: PathBuf) -> Self {
-        UserCrate(StateBuilt::new(db_oid, fn_oid, shared_object))
+    pub(crate) fn built(db_oid: pg_sys::Oid, fn_oid: pg_sys::Oid, shared_object: &Path) -> Self {
+        UserCrate(StateBuilt::new(db_oid, fn_oid, shared_object.to_path_buf()))
```
workingjubilee, line 94: If we know we're just going to allocate anyways, then this change just prompts a reallocation, and it's reasonable to ask the caller to use `.to_owned()` on their `&Path`. This should remain as it was.

```suggestion
    pub(crate) fn built(db_oid: pg_sys::Oid, fn_oid: pg_sys::Oid, shared_object: PathBuf) -> Self {
        UserCrate(StateBuilt::new(db_oid, fn_oid, shared_object))
```

src/plrust_proc.rs
```diff
@@ -0,0 +1,117 @@
...
+    let temp_so_file = tempfile::Builder::new().tempfile_in(work_dir)?;
+    std::fs::write(&temp_so_file, so)?;
+
+    // SAFETY: Postgres globally sets this to `const InvalidOid`, so is always read-safe,
+    // then writes it only during initialization, so we should not be racing anyone.
+    let db_oid = unsafe { MyDatabaseId };
+
+    // fabricate a StateBuilt version of the UserCrate so that we can "load()" it -- tho we're
+    // long since past the idea of crates, but whatev, I just work here
+    let built = UserCrate::built(db_oid, pg_proc_oid, temp_so_file.path());
```
workingjubilee, line 89: As mentioned.
```suggestion
    let built = UserCrate::built(db_oid, pg_proc_oid, temp_so_file.path().to_owned());
```

src/plrust.rs
```diff
@@ -103,11 +80,20 @@ pub(crate) fn compile_function(fn_oid: pg_sys::Oid) -> eyre::Result<(PathBuf, Ou
...
 
-    Ok((shared_object.into(), output))
+    // store the shared object in our table
+    plrust_proc::create_or_replace_function(fn_oid, shared_object)?;
+
+    // cleanup after ourselves
+    tracing::trace!("removing {}", shared_object.display());
+    std::fs::remove_file(shared_object)?;
+    tracing::trace!("removing {}", crate_dir.display());
+    std::fs::remove_dir_all(crate_dir)?;
```
workingjubilee, line 94: I think we shouldn't simply `?` out here but rather we should prefer to proceed (thus ignoring the errors) if we can't clean up, but still report the error occurred in tracing. We "have what we want".

src/plrust.rs
```diff
@@ -93,7 +70,7 @@ pub(crate) unsafe fn evaluate_function(
 }
 
 #[tracing::instrument(level = "debug")]
-pub(crate) fn compile_function(fn_oid: pg_sys::Oid) -> eyre::Result<(PathBuf, Output)> {
+pub(crate) fn compile_function(fn_oid: pg_sys::Oid) -> eyre::Result<Output> {
```
workingjubilee, line 73: Hm. Were we really ignoring this PathBuf at every call site? At first it feels weird but I guess it's fair to clean it up, yeah.

build.rs
```diff
@@ -0,0 +1,7 @@
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+    println!(
+        "cargo:rustc-env=TARGET={}",
+        std::env::var("TARGET").unwrap()
+    );
+}
```
workingjubilee, line 7: whyfor build.rs?

src/plrust_proc.rs
```diff
@@ -0,0 +1,117 @@
...
+    // need it.  If any of the above failed and returned an Error, it'll still get deleted when
+    // the function returns.
+    drop(temp_so_file);
+
+    // all good
+    Ok(loaded)
+}
+
+#[inline]
+fn pkey_datums(pg_proc_oid: pg_sys::Oid) -> Vec<(PgOid, Option<pg_sys::Datum>)> {
```
workingjubilee, line 102: comment explaining this?

workingjubilee (changes requested):
Most of this looks good. I need to reread hooks.rs a second time, but I had some comments that you can address nowish.

README.md
```diff
@@ -24,14 +24,22 @@ sum_array
...
+# Configuration
 
-There are two `postgresql.conf` settings that must be configured:
+First, plrust must be configured as a `shared_preload_libraries` entry in `postgresql.conf`.  For example:
+
+```
+shared_preload_libraries = 'plrust'
+```
+
+Failure to do so will cause unexpected behavior around the DROP FUNCTION, DROP SCHEMA, and ALTER FUNCTION commands.
```
workingjubilee, line 35: I feel like we could have a script that people can run that preloads the required settings, but then they also have to manually set the `plrust.pg_config` and `plrust.work_dir` anyways...

src/lib.rs
```diff
@@ -37,6 +40,12 @@ use pgx::{pg_getarg, prelude::*};
 pub use tests::pg_test;
 pgx::pg_module_magic!();
 
+/// this function will raise an ERROR if the `plrust` language isn't installed
+pub(crate) fn plrust_lang_oid() -> Option<pg_sys::Oid> {
+    static PLRUST_LANG_NAME: &[u8] = b"plrust\0"; // want this to look like a c string
+    Some(unsafe { pg_sys::get_language_oid(PLRUST_LANG_NAME.as_ptr().cast(), false) })
```
workingjubilee, line 46: Hm. Why do we return `Option` here, in light of this? Just to make forward compatibility easier?
```suggestion
    // SAFETY: We pass `missing_ok: false`, which will cause an error if not found.
    // So we always will return a valid Oid if we return at all.
    // The first parameter has the same requirements as `&CStr`.
    Some(unsafe { pg_sys::get_language_oid(PLRUST_LANG_NAME.as_ptr().cast(), false) })
```

build.rs
eeeebbbbrrrr: To get the target triple available at runtime. Suggested by @thomcc

workingjubilee: Hmm, wouldn't we want this `build.rs` instead on the user fn when we compile them?

eeeebbbbrrrr: No.  It's the plrust extension itself that needs to know this information as it's used to decide if we have a function.so compiled for its host.  The function.so itself doesn't care.

README.md
eeeebbbbrrrr: I don't think we could come up with a script that could guess any of this for any given installation, and that's most definitely not common practice in the postgres extension world.

src/hooks.rs
```diff
@@ -0,0 +1,264 @@
...
+    if lang_oid == plrust_lang_oid() {
+        // block a change to the 'STRICT' property
+        let actions = PgList::<pg_sys::DefElem>::from_pg(alter_stmt.actions);
+        for defelem in actions.iter_ptr() {
+            static STRICT: &str = "strict";
+
+            let defelem = PgBox::from_pg(defelem);
+            let name = CStr::from_ptr(defelem.defname);
+
+            if name.to_string_lossy().eq_ignore_ascii_case(STRICT) {
```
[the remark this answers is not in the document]
workingjubilee: I guess we're iterating over `defelem`s so I... are those guaranteed to be individual properties?

src/plrust.rs
eeeebbbbrrrr: I'm not convinced that's a good idea.  We're running inside a database that likes to fail fast when something isn't as expected.  Not being able to delete a file we just created would most surely indicate some kind of terrible problem.  Waving that away doesn't seem like a nice thing to do for the user or for Postgres.

I get that it's "okay" as far as we're concerned, but I don't feel like plrust is a good arbiter of making these kinds of decisions for postgres or users.

src/plrust.rs
eeeebbbbrrrr: We weren't, but we are now since the function is stored in the database.

build.rs
workingjubilee: oh so the same reason as [`current_platform::CURRENT_PLATFORM`](https://github.com/tcdi/plrust/blob/b8ffb96eabb758b03e3ec2d07c001627ba21332f/src/plrust.rs#L24) here

src/hooks.rs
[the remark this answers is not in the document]
eeeebbbbrrrr: I argued with myself around this.  Sadly Postgres doesn't have any kind of enum for these things.  No matter how the user spells it and no matter surrounding whitespace, Postgres' parser is always going to go for `"strict"`, but I thought a little bit of fuzziness was acceptable considering how terrible it would be for this switch to be flipped.

We could to something like `.to_lowercase().contains(STRICT)` but then Postgres 16 is gonna invent some new function probably that has the word "strict" in it that has nothing to do with this.  I mean, they probably won't, but...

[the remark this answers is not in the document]
eeeebbbbrrrr: also, each "other parameter" is its own `DefElem`, so we at least have that going for us.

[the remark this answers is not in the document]
workingjubilee: > We could to something like .to_lowercase().contains(STRICT) but then Postgres 16 is gonna invent some new function probably that has the word "strict" in it that has nothing to do with this. I mean, they probably won't, but...

I am 100% okay with breaking on attempting to ALTER FUNCTION that.

src/hooks.rs
```diff
@@ -0,0 +1,264 @@
+use std::ffi::CStr;
```
workingjubilee, line 1: I am okay if you then wrap every single `unsafe fn`'s internals in a single giant `unsafe {}` block with a brief comment on what's going on that makes it internally `unsafe` as well, but I would prefer to be able to see where the `unsafe {}` is. If any of the existing `unsafe fn` are not actually `unsafe`, then I would like to see that too.
```suggestion
#![deny(unsafe_op_in_unsafe_fn)]
use std::ffi::CStr;
```

workingjubilee (changes requested):
The main reason that `#![deny(unsafe_op_in_unsafe_fn)]` isn't crate-wide yet is because the symbol loading code is subtle and needs careful auditing that I haven't gotten around to yet. I can approve if this step is done and I can see the result.

src/plrust_proc.rs
workingjubilee: Nice, thank you.

src/plrust.rs
workingjubilee: Hmm, okay. In that case, maybe we should annotate these with `.wrap_err("failed during function cleanup")?` to provide context on the error?

src/hooks.rs
eeeebbbbrrrr: I don't follow?  Is `std::ffi::CStr` some kind of problem?

build.rs
eeeebbbbrrrr: I dunno.  There's no explanation as to why that's there.

JohnHVancouver: It's used to include the target platform in the name so it'll fail if the .so was compiled for a different architecture, so effectively the same thing

README.md
JohnHVancouver, line 35: Do you want to add something like `process_shared_preload_libraries_in_progress` check?
https://github.com/pgaudit/pgaudit/blob/master/pgaudit.c

It'll only error during CREATE LANGUAGE/CREATE EXTENSION plrust is called but that seems better than nothing

src/hooks.rs
```diff
@@ -0,0 +1,264 @@
...
+        }
+    } else if utility_stmt.type_ == pg_sys::NodeTag_T_AlterFunctionStmt {
+        let alter_stmt = PgBox::from_pg(pstmt.utilityStmt.cast::<pg_sys::AlterFunctionStmt>());
+
+        // and for AlterFunctioStmt, we'll just go do it.
+        handle_alter_function(&alter_stmt);
+    }
+
+    if PREVIOUS_PROCESS_UTILITY_HOOK.is_some() {
+        // previous hook must go last as we need the catalog entries this utility statement might
```
JohnHVancouver, line 107: Is your concern here that other hooks might maul the catalog entries we need here?

src/plrust_proc.rs
```diff
@@ -0,0 +1,119 @@
...
+use crate::gucs;
+use crate::user_crate::{StateLoaded, UserCrate};
+use pgx::pg_sys::MyDatabaseId;
+use pgx::{extension_sql, pg_sys, IntoDatum, PgBuiltInOids, PgOid, Spi};
+use std::path::Path;
+
+extension_sql!(
+    r#"
+CREATE TABLE plrust.plrust_proc (
+    id            regproc   NOT NULL,
```
JohnHVancouver, line 12: Ah I remember why I didn't use regproc and decided to go with object identity:

https://www.postgresql.org/docs/current/pgupgrade.html

pg_upgrade doesn't allow you to upgrade with regproc. 

Which also brings up an interesting question of how does this work with pg_upgrade...

If you use something like *link* mode, the data files including this table will be copied over? But CREATE OR REPLACE FUNCTION would run as part of pg_restore iirc, so you'll have stale entries in there presumably since I don't think functionOids are preserved right now

build.rs
eeeebbbbrrrr: I'm removing the dependency on that crate in favor of our own build.rs.  Today maybe they're basically the same, but down the road I expect plrust's "target_triple" to become something much more complex and I'd prefer us to just go ahead and make the decision to control it

src/hooks.rs
eeeebbbbrrrr: We talked about this offline and I understand.  Will do.

README.md
eeeebbbbrrrr: Oh, yes.  That's great.  Didn't know was a thing.

src/hooks.rs
eeeebbbbrrrr: I suppose you could argue we ought to go last in stead of first, in case some other extension tries to change the "STRICT" property of a plrust function.  But...

The whole hook system is a farce anyways, as changing the listed order in `shared_preload_libraries` will change what the "previous hook" is.  And it requires that extension to even bother to call the previous hook, which isn't enforced in any way.

src/plrust_proc.rs
eeeebbbbrrrr: I'll have to think about this, but right now my gut reaction is that this is pg_upgrade's problem, not pl/rust's.

src/hooks.rs
eeeebbbbrrrr: Thinking about this, I think we want to go before the previous hook for DROP FUNCTION/SCHEMA, but go after the hook for ALTER FUNCTION.

That seems like the right thing to do.  I'll rejigger the code.

eeeebbbbrrrr:
Okay, I'm done with this PR.  I think I've addressed everything to the best of my abilities.  Hoping to see it merged Monday as I'm close to putting up another PR that is based on this one.

src/plrust_proc.rs
JohnHVancouver: Yea but if we don't support it, I think it makes PL/Rust un-useable by anyone who cares about migrating to latest versions of Postgres which is going to be almost everyone eventually.

I'm skeptical the community will relax it's restrictions on regproc anytime soon.

It'll be a hindrance to adoption, and more of an unpleasant surprise when folks try upgrading since I doubt anyone actually reads the documentation properly.

IIUC you're just looking for a unique identity which you should be able to get by just using:

https://www.postgresql.org/docs/9.4/functions-info.html
`pg_identify_object`

I think it's a bit more of a hassle if the function schema migrates I guess but that should be able to be handled

src/hooks.rs
JohnHVancouver: No I think you're right. The API isn't the greatest. We can do "best effort" but since we're dependent on ordering and what the previous hook decided to do it's all up in the air

src/plrust_proc.rs
eeeebbbbrrrr: As I understand it, pg_upgrade will migrate our table with the existing regproc values.  Then at some point it'll CREATE OR REPLACE all user functions.  Then our table will have the old values, which are now likely stale, and the entries for the new CREATE OR REPLACE execution.

Other than some garbage in our table, I don't see the problem.  We can expose a function to garbage collect it -- it's a simple DELETE statement.

workingjubilee:
Thank you so much!

workingjubilee (approved):

src/plrust_proc.rs
JohnHVancouver: It won't even try migrating. It checks if regproc is being used and blocks the upgrade. 
https://github.com/postgres/postgres/blob/master/src/bin/pg_upgrade/check.c#L1273

```
pg_upgrade does not support upgrading of databases containing table columns using these reg* OID-referencing system data types:
```

https://www.postgresql.org/docs/current/pgupgrade.html

Here's a sample repro I just did


```
postgres=> create table test (id regproc);
CREATE TABLE
postgres=> insert into test values (16405);
INSERT 0 1
postgres=> select * from test;
    id     
-----------
 increment
 ````
 
 Trigger pg_upgrade....
 
```
Your installation contains one of the reg* data types in user tables.
These data types reference system OIDs that are not preserved by
pg_upgrade, so this cluster cannot currently be upgraded.  You can
drop the problem columns and restart the upgrade.
A list of the problem columns is in the file:
    tables_using_reg.txt
```

I agree the case you're doing *could* be supported in theory since the handler handles all of that but it just blocks cause funcOid's aren't preserved so for the sake of data integrity they just prevent it.

I actually don't know if this scheme even will work with pg_upgrade if you're putting it in a table now that I think about it....

https://github.com/postgres/postgres/blob/master/src/bin/pg_upgrade/pg_upgrade.c#L159
https://github.com/postgres/postgres/blob/master/src/bin/pg_upgrade/pg_upgrade.c#L169-L172

We run pg_restore in `create_new_objects` which I'm not sure if it'll recreate the extension table at that step or not. And if it does and you run link it would break. 

I think that's more reason to use objectIdentity actually, cause in CREATE FUNCTION, you could just avoid recompiling if you know this is during pg_upgrade and that would be faster and you would get database link for free without needing to redo anything

eeeebbbbrrrr: You're harshing my mellow, man.

I'll take a look at this today/tomorrow and put up a PR.  It won't be a big deal to use identify_object.  Will just need to make sure it's out of the "hot path" for execution.

Pretty annoying pg_upgrade can't handle this.

eeeebbbbrrrr: > You're harshing my mellow, man.

Having to also deal with the function schema changing is annoying too.  We already have the hook in place so the hard work has been done, it's just now more complexity.

eeeebbbbrrrr: > Having to also deal with the function schema changing is annoying too

Which, of course, also means on rename of a schema that contains plrust functions.

Geez, anything else I'm not thinking of?

JohnHVancouver: No I think that's about it. Off the top of my head it should just be:

1. Prevent schema from being renamed
2. Prevent direct table access for INSERT/DELETE? (SELECT seems fine)

PR #124, opened 2022-11-21
Reload function.so when the function changes

eeeebbbbrrrr:
This PR, which is based on the changes in #122, reloads a function if it was changed by a concurrent transaction.  

It also cleans up our various syscache usages and centralizes them into a new type named `PgProc` in the `pgproc.rs` module.

Note, it'll be easier to review this after #122 is merged.

### DIFF
--- a/src/pgproc.rs
+++ b/src/pgproc.rs
@@ -1,0 +1,96 @@
+use pgx::{pg_sys, FromDatum, IntoDatum};
+use std::mem::ManuallyDrop;
+use std::ptr::NonNull;
+
+/// Provides a safe wrapper around a Postgres "SysCache" entry from `pg_catalog.pg_proc`.
+pub(crate) struct PgProc {
+    inner: ManuallyDrop<NonNull<pg_sys::HeapTupleData>>,
+}
+
+impl Drop for PgProc {
+    fn drop(&mut self) {
+        unsafe { pg_sys::ReleaseSysCache(self.inner.assume_init().as_ptr()) }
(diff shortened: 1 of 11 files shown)


src/pgproc.rs
```diff
@@ -0,0 +1,98 @@
+use pgx::{pg_sys, FromDatum, IntoDatum};
+use std::mem::ManuallyDrop;
+use std::ptr::NonNull;
+
+/// Provides a safe wrapper around a Postgres "SysCache" entry from `pg_catalog.pg_proc`.
+pub(crate) struct PgProc {
+    inner: ManuallyDrop<NonNull<pg_sys::HeapTupleData>>,
```
eeeebbbbrrrr, line 7: This is the only thing I did that I ain't too sure about.  Dunno if ManuallyDrop is required here since below on line 10 Drop is implemented for `PgProc`.

workingjubilee: Last I checked, Rust doesn't attempt to run Drop on things behind pointers, so ManuallyDrop is indeed redundant here (but if there's a need to carefully handle HeapTupleData's fields you should note that pervasively, obv).

eeeebbbbrrrr: Not the fields per se but the pointer does need to be released.  I’ll circle back and clean this up. Don’t think it’s worth holding up a merge. Thanks!

workingjubilee (approved):
Yeah, this looks much nicer. Thank you!

PR #128, opened 2022-12-01
Catch `#[no_mangle]`

workingjubilee:
This PR adds code sufficient to catch the usage of `#[no_mangle]`. Note that deploying PL/Rust with the given Rust toolchain currently indicated by our `rust-toolchain.toml` will not in fact catch `#[link_section]`, despite the branch name. That requires an upgrade to a more recent version of Rust, which is incompatible with use with `postgrestd`.

### DIFF
--- a/src/plrust.rs
+++ b/src/plrust.rs
@@ -107,7 +107,8 @@ pub(crate) fn compile_function(fn_oid: pg_sys::Oid) -> eyre::Result<Output> {
     let provisioned = generated.provision(&work_dir)?;
     // We want to introduce validation here.
     let crate_dir = provisioned.crate_dir().to_path_buf();
-    let (built, output) = provisioned.build(pg_config, target_dir.as_path())?;
+    let (validated, output) = provisioned.validate(pg_config, target_dir.as_path())?;;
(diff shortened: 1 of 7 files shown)


eeeebbbbrrrr (changes requested):
I think in general this is fine, code wise.  It definitely needs some code comments/commentary about what's happening, why it works, the possible problems it prevents, etc, etc.

eeeebbbbrrrr (approved):
ship it!

eeeebbbbrrrr:
Thanks for the clarifications offline, @workingjubilee.  I know I can be dense sometimes.  I'd like to enshrine them here:

```@workingjubilee
So PGX already makes every function into function and function_wrapper
This is usually elided in common thought and usage of the library, but here we peel that curtain back. We don't need function_wrapper yet, and it's the "inherently unsafe" part of PGX code. 
So we write the library as if it only had whatever code the user function contains, which is not actually enough to make it work, but it is still Rust, and it still contains "whatever the function does".
Then we check that.
Then rewrite it using the full text, including the macro annotations that create function_wrapper 
Then build that.
This is valid because PGX's macros don't substantively change the function's code but rather most of what they do is creating the wrapper.
```

PR #134, opened 2022-12-09
Remove geiger

workingjubilee:


plrust/Cargo.toml
```diff
@@ -23,7 +23,7 @@ sandboxed = []
 # Forcibly enable a feature used by x86_64 MacOS machines because they're bad at `dlclose()`
 force_enable_x86_64_darwin_generations = []
 target_postgrestd = []
-verify = ["dep:geiger"]
+verify = []
```
eeeebbbbrrrr, line 26: Guess we don't need the feature anymore?

eeeebbbbrrrr (changes requested):

plrust/Cargo.toml
workingjubilee: > This leaves the "verify" feature even though it does nothing
as a note to myself in the future to hook more code into
allowing people to build a "plrustu".

I am happy to remove it ofc if you feel otherwise.

eeeebbbbrrrr: lets compromise on leaving it there, but commented out?  Since it's a public facing thing, I don't want it to imply that turning it on does something we don't actually do yet.

workingjubilee: Sure!

PR #140, opened 2022-12-13
use pg_identify_object instead of regproc for plrust.plrust_proc

JohnHVancouver:
regproc is not supported in pg_upgrade so use pg_identify_object to generate a unique identity for the function

Discussed using object identity in #122

### DIFF
--- a/plrust/src/plrust_proc.rs
+++ b/plrust/src/plrust_proc.rs
@@ -10,7 +10,7 @@ use std::path::Path;
 extension_sql!(
     r#"
 CREATE TABLE plrust.plrust_proc (
-    id            regproc   NOT NULL,
+    id            text   NOT NULL, -- pg_identify_object identity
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
I’ve already done this.

eeeebbbbrrrr:
There's a branch for it here:  https://github.com/tcdi/plrust/compare/main...pg_identify_object

I have not made a PR yet because it needs to be rebased

eeeebbbbrrrr:
@JohnHVancouver I'm closing this one in favor or #141 that I just put up.  It doesn't use SQL to get the identifier value, which seems fragile and slow.

JohnHVancouver:
Ah sorry didn't realize, thanks!

eeeebbbbrrrr:
NP.  It was my bad for not even putting up the PR.  I'm juggling a lot of chainsaws...

PR #141, opened 2022-12-13
use `pg_identify_object()` value as the pkey for `plrust.plrust_proc`

eeeebbbbrrrr:
See the comments that @JohnHVancouver made on his PR #140, but this is the right way to do this.  Doesn't require going through SPI to lookup the object identifier.

### DIFF
--- a/plrust/src/plrust_proc.rs
+++ b/plrust/src/plrust_proc.rs
@@ -140,7 +140,8 @@ fn get_fn_identity_datum(pg_proc_oid: pg_sys::Oid) -> (PgOid, Option<pg_sys::Dat
     let result = (PgBuiltInOids::TEXTOID.oid(), identity_str.into_datum());
 
     unsafe {
-        // SAFETY: identity_ptr was previously proven valid
+        // SAFETY: identity_ptr was previously proven valid and
+        // identity_str was reallocated elsewhere in Postgres
         pg_sys::pfree(identity_ptr.cast());
     }
 
--- a/plrust/src/plrust_proc.rs
+++ b/plrust/src/plrust_proc.rs
@@ -111,7 +111,7 @@ fn pkey_datums(pg_proc_oid: pg_sys::Oid) -> Vec<(PgOid, Option<pg_sys::Datum>)> 
 // helper function to build the function identity (oid, value) datum
 fn get_fn_identity_datum(pg_proc_oid: pg_sys::Oid) -> (PgOid, Option<pg_sys::Datum>) {
     let oa = pg_sys::ObjectAddress {
-        classId: pg_sys::ProcedureRelationId,
+        classId: pg_sys::ProcedureRelationId, // the "oid" of Postgres' `pg_catalog.pg_proc` table
         objectId: pg_proc_oid,
         objectSubId: 0,
     };
--- a/plrust/src/plrust_proc.rs
+++ b/plrust/src/plrust_proc.rs
@@ -118,15 +118,15 @@ fn get_fn_identity_datum(pg_proc_oid: pg_sys::Oid) -> (PgOid, Option<pg_sys::Dat
     let identity_ptr = unsafe {
         #[cfg(feature = "pg13")]
         {
-            // SAFETY:  getObjectIdentity will raise an ERROR if theObjectAddress we created doesn't
-            // exist, otherwise returning a properly palloc'd pointer
+            // SAFETY:  getObjectIdentity will raise an ERROR if the ObjectAddress we created doesn't
+            // exist, otherwise it returns a properly palloc'd pointer
             pg_sys::getObjectIdentity(&oa as *const _)
         }
 
         #[cfg(not(feature = "pg13"))]
         {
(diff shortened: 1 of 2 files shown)


plrust/src/plrust_proc.rs
```diff
@@ -105,19 +99,55 @@ pub(crate) fn load(pg_proc_oid: pg_sys::Oid) -> eyre::Result<UserCrate<StateLoad
...
 
+// helper function to build the function identity (oid, value) datum
+fn get_fn_identity_datum(pg_proc_oid: pg_sys::Oid) -> (PgOid, Option<pg_sys::Datum>) {
+    let oa = pg_sys::ObjectAddress {
+        classId: pg_sys::ProcedureRelationId,
+        objectId: pg_proc_oid,
+        objectSubId: 0,
+    };
+    let identity_ptr = unsafe {
+        #[cfg(feature = "pg13")]
```
JohnHVancouver, line 120: I assume this should be <=13 here, not sure if that's possible? Or does PL/Rust only support >= 13

plrust/src/plrust_proc.rs
```diff
@@ -105,19 +99,55 @@ pub(crate) fn load(pg_proc_oid: pg_sys::Oid) -> eyre::Result<UserCrate<StateLoad
...
+    let identity_str = unsafe {
+        // SAFETY:  Postgres has given us a valid, albeit palloc'd, cstring as the result of getObjectIdentity
+        CStr::from_ptr(identity_ptr).to_str().unwrap_or_else(|_| {
+            pgx::error!("function {pg_proc_oid}'s identity is not a valid UTF8 string")
+        })
+    };
+
+    let result = (PgBuiltInOids::TEXTOID.oid(), identity_str.into_datum());
+
+    unsafe {
```
JohnHVancouver, line 143: I'm not familiar enough with pgx, but do we need to explicitly free palloc-d memory?

plrust/src/plrust_proc.rs
```diff
@@ -105,19 +99,55 @@ pub(crate) fn load(pg_proc_oid: pg_sys::Oid) -> eyre::Result<UserCrate<StateLoad
...
-            get_target_triple().into_datum(),
-        ),
+        get_fn_identity_datum(pg_proc_oid),
+        (PgBuiltInOids::TEXTOID.oid(), get_target_triple().into_datum()),
     ]
 }
 
+// helper function to build the function identity (oid, value) datum
+fn get_fn_identity_datum(pg_proc_oid: pg_sys::Oid) -> (PgOid, Option<pg_sys::Datum>) {
+    let oa = pg_sys::ObjectAddress {
```
JohnHVancouver, line 114: Ty this is a lot cleaner

plrust/src/plrust_proc.rs
eeeebbbbrrrr: Only >=13.

plrust/src/plrust_proc.rs
eeeebbbbrrrr: Are you asking why we pfree it?  Because it would leak until CurrentMemoryContext is free'd.  Which could be okay, I suppose.

JohnHVancouver: Yup, ok that seems harmless to do earlier then.

JohnHVancouver (approved):

plrust/src/plrust_proc.rs
```diff
@@ -105,19 +99,55 @@ pub(crate) fn load(pg_proc_oid: pg_sys::Oid) -> eyre::Result<UserCrate<StateLoad
...
+            pgx::error!("function {pg_proc_oid}'s identity is not a valid UTF8 string")
+        })
+    };
+
+    let result = (PgBuiltInOids::TEXTOID.oid(), identity_str.into_datum());
+
+    unsafe {
+        // SAFETY: identity_ptr was previously proven valid
+        pg_sys::pfree(identity_ptr.cast());
+    }
```
workingjubilee, line 146: I think this has another hidden assumption
```suggestion
    unsafe {
        // SAFETY: identity_ptr was previously proven valid and
        // identity_str was reallocated elsewhere in Postgres
        pg_sys::pfree(identity_ptr.cast());
    }
```

plrust/src/plrust_proc.rs
```diff
@@ -105,19 +99,55 @@ pub(crate) fn load(pg_proc_oid: pg_sys::Oid) -> eyre::Result<UserCrate<StateLoad
...
-        ),
+        get_fn_identity_datum(pg_proc_oid),
+        (PgBuiltInOids::TEXTOID.oid(), get_target_triple().into_datum()),
     ]
 }
 
+// helper function to build the function identity (oid, value) datum
+fn get_fn_identity_datum(pg_proc_oid: pg_sys::Oid) -> (PgOid, Option<pg_sys::Datum>) {
+    let oa = pg_sys::ObjectAddress {
+        classId: pg_sys::ProcedureRelationId,
```
workingjubilee, line 115: Hm. A constant? What does this describe?

eeeebbbbrrrr:
I approve this PR

PR #142, opened 2022-12-13
Introduce a `trusted-pgx` module

eeeebbbbrrrr:
Add a new module, `trusted-pgx` which exports the bare minimum things necessary for a standard plrust user function to not only compile, but to also pass safety checks, and to otherwise be usable to the developer.

### Notes

- We have a chicken/egg problem right now as the `trusted-pgx` crate isn't published, which is why its dependency is hardcoded to this repo + branch

- The user crate dependency on `pallocator` has been commented out.  Including it causes compilation errors

- Unclear if `trusted-pgx` exports everything a plrust user function will need.  I think it's probably pretty close

- All tests pass locally

### DIFF
--- a/plrust/src/user_crate/crating.rs
+++ b/plrust/src/user_crate/crating.rs
@@ -149,32 +149,7 @@ impl FnCrating {
             "Generating `Cargo.toml`"
         );
 
-        let cargo_toml = toml::toml! {
-                    [package]
-                    edition = "2021"
-                    name = crate_name
-                    version = "0.0.0"
-
-                    [features]
-                    default = [version_feature]
-
-                    [lib]
-                    crate-type = ["cdylib"]
-
-                    [dependencies]
-                    pgx =  { git = "https://github.com/tcdi/pgx", branch = "develop" }
-                    // pallocator = { version = "0.1.0", git = "https://github.com/tcdi/postgrestd", branch = "1.61" }
-
-                    /* User deps added here */
-
-                    [profile.release]
-                    debug-assertions = true
-                    codegen-units = 1_usize
-                    lto = "fat"
-                    opt-level = 3_usize
-                    panic = "unwind"
-        };
-
+        let cargo_toml = cargo_toml_template(&crate_name, &version_feature);
         match cargo_toml {
             toml::Value::Table(mut cargo_manifest) => {
                 // We have to add the user deps now before we return it.
@@ -289,8 +264,39 @@ fn compose_lib_from_mods<const N: usize>(modules: [syn::ItemMod; N]) -> eyre::Re
 /// Used by both the unsafe and safe module.
 pub(crate) fn shared_imports() -> syn::ItemUse {
     syn::parse_quote!(
+        // we (plrust + pgx) fully qualify all pgx imports with `::pgx`, so if the user's function
+        // doesn't use any other pgx items we don't want a compiler warning
+        #[allow(unused_imports)]
         use pgx::prelude::*;
     )
+}
+
+pub(crate) fn cargo_toml_template(crate_name: &str, version_feature: &str) -> toml::Value {
+    toml::toml! {
+        [package]
+        edition = "2021"
+        name = crate_name
+        version = "0.0.0"
+
+        [features]
+        default = [version_feature]
+
+        [lib]
+        crate-type = ["cdylib"]
+
+        [dependencies]
+        pgx =  { git = "https://github.com/tcdi/plrust/trusted-pgx", branch = "trusted-pgx", package = "trusted-pgx" }
+        // pallocator = { version = "0.1.0", git = "https://github.com/tcdi/postgrestd", branch = "1.61" }
+
+        /* User deps added here */
+
+        [profile.release]
+        debug-assertions = true
+        codegen-units = 1_usize
+        lto = "fat"
+        opt-level = 3_usize
+        panic = "unwind"
+    }
 }
 
 fn unsafe_mod(mut called_fn: syn::ItemFn, variant: &CrateVariant) -> eyre::Result<syn::ItemMod> {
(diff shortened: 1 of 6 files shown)


eeeebbbbrrrr:
Question:  Should the `trusted-pgx` crate live here or should it be part of `pgx`?  I can be persuaded either way.  

It definitely needs to get out on crates.io and it is `pgx` that does the publishing.

eeeebbbbrrrr:
Worth nothing that this **drastically** improves compilation times and also reduces the resulting .so size a bit:

```sql
create or replace function one_rust() 
returns int 
language plrust as $$ 
      error!("hi"); 
      Some(1) 
$$;
```

### clean

- main: 01m 31.198s
- trusted-pgx: 29.867s

### dirty

- main: 6.244s
- trusted-pgx: 3.199s

### .so size

- main: 400,164
- trusted-pgx: 357,076

trusted-pgx/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+crate-type = ["rlib"]
+
+[features]
+pg13 = ["pgx/pg13"]
+pg14 = ["pgx/pg14"]
+pg15 = ["pgx/pg15"]
+
+[dependencies]
+pgx = { git = "https://github.com/tcdi/pgx", branch = "develop", features = [ "no-schema-generation", "plrust" ], default-features = false }
+#pgx = { path = "/Users/e_ridge/_work/pgx/pgx", features = [ "no-schema-generation", "plrust" ], default-features = false }
```
JohnHVancouver, line 21: Can be removed?

trusted-pgx/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+pub use iter::*;
+pub mod iter {
+    pub use ::pgx::iter::{SetOfIterator, TableIterator};
+}
+
+pub use memcxt::*;
+pub mod memcxt {
+    pub use ::pgx::memcxt::PgMemoryContexts;
```
JohnHVancouver, line 38: I'm trying to reason if there's anything concerning with allowing regular users to manipulate memory contexts directly. Worst case they mess up their own allocations and drop potential allocations they might rely on later..? 

I guess it can be used to palloc large amounts of memory directly leading to stability issues?

plrust/src/user_crate/crating.rs
```diff
@@ -289,10 +264,41 @@ fn compose_lib_from_mods<const N: usize>(modules: [syn::ItemMod; N]) -> eyre::Re
...
+
+        [features]
+        default = [version_feature]
+
+        [lib]
+        crate-type = ["cdylib"]
+
+        [dependencies]
+        pgx =  { git = "https://github.com/tcdi/plrust", branch = "trusted-pgx", package = "trusted-pgx" }
+        // pallocator = { version = "0.1.0", git = "https://github.com/tcdi/postgrestd", branch = "1.61" }
```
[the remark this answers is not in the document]
eeeebbbbrrrr: I haven’t yet investigated why it gets angry if used. Task for tomorrow.

trusted-pgx/Cargo.toml
eeeebbbbrrrr: Yeah, before merge. Makes it easy for me to mess with things locally in the mean time.

trusted-pgx/src/lib.rs
eeeebbbbrrrr: It’s required for pgx in general. 

We can audit PgMemoryContexts and make sure the things that should be “unsafe” are. And if that means every function that’s probably okay. 

Or I can implement a pared down version just for trusted-pgx.

workingjubilee: The latter already can be done via PL/Perl: simply make a bunch of arrays and fill them with meaningless junk data. I expect the same is true of PL/pgSQL though it might require something slightly more clever due to it being more heavily query-oriented.

Another good "joke" in that vein, this time in Rust, is to try `vec![1; 4_294_967_295]`.

Mind, any functions that allow them to actually violate memory safety *otherwise* should be marked `unsafe`.

eeeebbbbrrrr: > Mind, any functions that allow them to actually violate memory safety otherwise should be marked unsafe.

yup. That’s really plrust’s only concern

eeeebbbbrrrr: I'm working on this over on https://github.com/tcdi/pgx/pull/995.  That should get merged into pgx's develop branch later today, and then plrust's "unsafe" guards will simply prevent users from using PgMemoryContexts, and the other things in that PR.

plrust/src/user_crate/mod.rs
```diff
@@ -456,30 +456,8 @@ mod tests {
...
-                /* User deps added here */
-
-                [profile.release]
-                debug-assertions = true
-                codegen-units = 1_usize
-                lto = "fat"
-                opt-level = 3_usize
-                panic = "unwind"
-            };
+            let fixture_cargo_toml = cargo_toml_template(&crate_name, &version_feature);
```
workingjubilee, line 459: This seems like a reasonable answer, yeah. I kind of have grown to appreciate how the tests can be annoyingly, nitpickingly exacting, especially when modifying how the crate expands into Rust code, but this particular repetition is fruitless.

plrust/src/user_crate/crating.rs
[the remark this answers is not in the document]
workingjubilee: It's usually just a mismatching dependency. Best to sort it out after the pgx version bump per se.

workingjubilee (approved):
Neat! This looks like it'll be fine with the final cleanup pass in. If you need to leave the pallocator crate issue for me to kick into place, I can make that happen. Otherwise I will probably look at other ways to fortify this whole contraption.

plrust/src/user_crate/crating.rs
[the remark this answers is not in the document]
workingjubilee: I did some research and it is very much a thing to fix after pgx's version bump.

PR #143, opened 2022-12-14
Diet of WORMs

workingjubilee:
This allows PL/Rust to feed on a filesystem that has been configured to only allow "write-once read-many" or WORM. It fixes https://github.com/tcdi/plrust/issues/139 as one might imagine.

The unusual path remapping hack is only for this PR, I intend to fix it in the next. For various reasons I don't want to chance losing the git history for these files (yet).

### DIFF
--- a/plrust/src/plrust.rs
+++ b/plrust/src/plrust.rs
@@ -9,7 +9,7 @@ Use of this source code is governed by the PostgreSQL license that can be found 
 
 use crate::{
     gucs, plrust_proc,
-    user_crate::{StateLoaded, UserCrate},
+    user_crate::{FnReady, UserCrate},
 };
 
 use pgx::{pg_sys::FunctionCallInfo, pg_sys::MyDatabaseId, prelude::*};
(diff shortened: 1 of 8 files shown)


eeeebbbbrrrr:
Probably the only thing is I feel all the `user_crate/*.rs` are misnamed now that the state names have changed.

As an aside, I've **never** liked how some tests work.  Comparing generated v/s expected rust code using syn/quote causes us to write that generated code multiple times.  And it make reviewing a diff like this one incredibly difficult.  Doesn't have to be part of this PR, but is there something we can do about this?

workingjubilee:
Yes, I want to do the `git mv` steps as something that gets merged as a separate commit so the history still follows.

Yeah, I made the tests use a more parametric form (partially generated, checking only the expected outline) to minimize that. I feel there's a lot of value in checking the overall outline against the fixture, and not as much in checking that every line is exactly the same, so I'm hoping the new form will tend to resist breaking on small changes.

eeeebbbbrrrr:
> Yes, I want to do the `git mv` steps as something that gets merged as a separate commit so the history still follows.

Okay.  I'll approve this now knowing that'll follow.

> Yeah, I made the tests use a more parametric form (partially generated, checking only the expected outline) to minimize that. I feel there's a lot of value in checking the overall outline against the fixture, and not as much in checking that every line is exactly the same, so I'm hoping the new form will tend to resist breaking on small changes.

I'm not sure there is.  These code checking tests only validate we typed the same sequence of characters in different parts of our code.  They don't ensure the character sequences are correct as it relates to plrust's promises of "safety" and "trust".

I don't have a better idea at the moment, so just take that for what it's worth right now.

eeeebbbbrrrr (approved):

Issue #78, opened 2022-09-12
BorrowMutError when a function errors once


JohnHVancouver:
A BorrowMutError occurs if a function errored once previously preventing it from being executed. This would prevent more complex functions from being ran again in the same session once it's errored.

It occurs here https://github.com/tcdi/plrust/blob/main/src/plrust.rs#L50
`loaded_symbols` is still borrowed after an error. Related to error handling in pgx?

```
plrust=# CREATE OR REPLACE FUNCTION raise_error() RETURNS TEXT
    IMMUTABLE STRICT
    LANGUAGE PLRUST AS
$$
[dependencies]
    # Add Cargo.toml dependencies here.
    
[code]
    
                                                                   
    pgx::error!("foo error");
    Some("hi".to_string())
$$;
CREATE FUNCTION
plrust=# select raise_error();
ERROR:  foo error
CONTEXT:  <unknown>:0:0
plrust=# select raise_error();
ERROR:  already borrowed: BorrowMutError
CONTEXT:  src/plrust.rs:50:56
plrust=# 
```

eeeebbbbrrrr:
I'm working on a fix for this.  It's real simple.  I think the problem in #79 is different tho.  That's related to recursion and actually trying to borrow the thing twice.  This is that we weren't properly protecting the FFI boundary when calling the user function in the loaded .so, so the stack didn't properly unwind.

PR #151, opened 2023-01-06
fix issue #78

eeeebbbbrrrr:
Properly protect the FFI boundary when calling the loaded user function so that the stack will properly unwind in the face of rust panics and Postgres errors.


workingjubilee (approved):
This looks good. I'll merge once CI passes and take a look at the recursive case.

Issue #79, opened 2022-09-13
BorrowMut error when nesting calls with SPI


jeff-davis:
```
create or replace function fn1(i int) returns int strict language plrust as $$
[code]
notice!("{}", "fn1 started");
let cmd = format!("select fn2({})", i);
Spi::execute(|client|
    {
        client.select(&cmd, None, None);
    });
notice!("{}", "fn1 finished");
Some(1)
$$;

create or replace function fn2(i int) returns int strict language plrust as $$
[code]
notice!("{}", "fn2 started");
notice!("{}", "fn2 finished");
Some(2)
$$;

select fn1(1);

NOTICE:  fn1 started
ERROR:  already borrowed: BorrowMutError
CONTEXT:  src/plrust.rs:50:56
SQL statement "select fn2(1)"
```

workingjubilee:
related to #78, maybe?

jeff-davis:
Likely, but I wasn't able to confirm so I filed a separate issue.

eeeebbbbrrrr:
Updated re-create:

```sql
create or replace function fn1(i int) returns int strict language plrust as $$
[code]
notice!("{}", "fn1 started");
let cmd = format!("select fn2({})", i);
Spi::connect(|client|
    {
        client.select(&cmd, None, None);
    });
notice!("{}", "fn1 finished");
Some(1)
$$;

create or replace function fn2(i int) returns int strict language plrust as $$
[code]
notice!("{}", "fn2 started");
notice!("{}", "fn2 finished");
Some(2)
$$;

select fn1(1);
```

This is **not** related to #78.  The problem is that `f1()` calls `f2()` through the executor and both are plrust functions, so we're recursing into `crate::plrust:::evaluate_function()`, and the `LOADED_SYMBOLS` static has already been borrowed by the first call to `f1()`.

We can either put `LOADED_SYMBOLS` behind a `Arc<RwLock>` or just go full unsafe and make it a `static mut: HashMap`.  I'm actually leaning towards the latter as Postgres isn't threaded and pgx otherwise prevents a user thread from doing FFI into postgres.

@workingjubilee, do you have any thoughts here?

workingjubilee:
[...]

PR #152, opened 2023-01-10
fix issue #79

eeeebbbbrrrr:
Recursively executing a plrust function is easy doable by putting the `UserCrate<FnReady>` behind a `Rc`.


workingjubilee (approved):
Neat! I figured it was probably the case that I was trying to fuss with the HashMap on the wrong nesting level, but I didn't realize it would be quite so easy this way. Cool.

PR #153, opened 2023-01-11
Fix up PL/Rust for `pg_sys::Oid` redesign

workingjubilee:
This also resolves some random compilation issues that ~~could hypothetically occur~~ will definitely occur until we actually cut the pgx release and bump.


eeeebbbbrrrr (approved):
Merging now.  Thanks!

PR #155, opened 2023-01-12
Work on cross-compilation support

eeeebbbbrrrr:
Add a GUC called `plrust.compilation_targets` that is a comma-separated list of other rust targets for which a plrust function should be compiled.  Update code to compile the plrust function for each target.

Also change from tracking file paths for the on-disk .so to its actual bytes as a `Vec<u8>`.  Also remove a bunch of unused getter functions from most of the states.


plrust/src/gucs.rs
```diff
@@ -91,3 +104,70 @@ pub(crate) fn tracing_level() -> tracing::Level {
         .map(|v| v.parse().expect("plrust.tracing_level was invalid"))
         .unwrap_or(tracing::Level::INFO)
 }
+
+#[derive(Debug, Clone, PartialOrd, PartialEq, Hash, Ord, Eq)]
+#[repr(transparent)]
+pub(crate) struct CompilationTarget(String);
```
workingjubilee, line 110: This sort of target-identification logic had previously been factored into its own file, and I would prefer it remain that way. Of course, if you have no preference, I can do the revision myself, as it's more important this get merged sometime soon so changes that follow on it can be done than that they are factored a particular way.

plrust/src/gucs.rs
```diff
@@ -65,6 +70,14 @@ pub(crate) fn init() {
         &PLRUST_ALLOWED_DEPENDENCIES,
         GucContext::Postmaster,
     );
+
+    GucRegistry::define_string_guc(
+        "plrust.compilation_targets",
+        "A comma-separated list of rust compilation 'target triples' to compile for",
+        "Useful for when it's known a system will replicate to a Postgres server on a different CPU architecutre",
```
workingjubilee, line 77:
```suggestion
        "Useful for when it's known a system will replicate to a Postgres server on a different CPU architecture",
```

eeeebbbbrrrr:
@workingjubilee I moved `src/user_crate/target.rs` up a level to just `src` and then moved that `CompilationTarget` struct into that, and adjusted the `target::triple()` function to use a `CompilationTarget` behind a `Lazy` so it only ever has to be calculated once.

Lemme know what you think.

I still need to address the things from #156 from @thomcc.

eeeebbbbrrrr:
@workingjubilee If this passes CI (and my smoke testing on my threadripper) we can go ahead and merge.  I know you wanted to take a look at removing the `PLRUST_TARGET` environment variable and were waiting on this.

Thoughts?

workingjubilee:
Yeah, I think that if this works we should merge it so followups can be parallelized.

eeeebbbbrrrr:
Works on my threadripper!  This is super cool.  Soon as CI passes I'll merge, assuming you don't have any other changes for me...

plrust/src/gucs.rs
eeeebbbbrrrr: I think I've addressed this in a good way.  Speak now or change it later!

eeeebbbbrrrr:
I'm going to assume the failure on pg15 is related to #160:  <https://github.com/tcdi/plrust/actions/runs/3916376672/jobs/6695427690>

eeeebbbbrrrr:
Trying this again.  Something is weird

plrust/src/gucs.rs
workingjubilee: Yeah, that seems good to me.

workingjubilee (approved):

PR #158, opened 2023-01-13
plrust functions now return `Result<Option<T>>`

eeeebbbbrrrr:
This does the work, which was mostly updating our tests, to change the wrapper function plrust writes for user functions to return `Result<Option<T>, Box<dyn Error>>` instead of a plain `Option<T>`.

This, I truly believe, is an ergonomics improvement now that `Spi` is based on returning Results.  And this is a thing we need to do know as it's a backwards incompatible ABI change along with a backwards incompatible code change.

### DIFF
--- a/plrust/src/tests.rs
+++ b/plrust/src/tests.rs
@@ -35,7 +35,7 @@ mod tests {
                 IMMUTABLE STRICT
                 LANGUAGE PLRUST AS
             $$
-                Some(a.into_iter().map(|v| v.unwrap_or_default()).sum())
+                Ok(Some(a.into_iter().map(|v| v.unwrap_or_default()).sum()))
             $$;
         "#;
         Spi::run(definition)?;
@@ -61,7 +61,7 @@ mod tests {
                 IMMUTABLE STRICT
                 LANGUAGE PLRUST AS
             $$
-                String::from("booper").into()
+                Ok(String::from("booper").into())
             $$;
         "#;
         Spi::run(definition)?;
@@ -78,7 +78,7 @@ mod tests {
                 IMMUTABLE STRICT
                 LANGUAGE PLRUST AS
             $$
-                String::from("swooper").into()
+                Ok(Some(String::from("swooper")))
             $$;
         "#;
         Spi::run(definition)?;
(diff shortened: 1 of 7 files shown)


plrust/src/tests.rs
```diff
@@ -264,7 +262,7 @@ mod tests {
                 IMMUTABLE STRICT
                 LANGUAGE PLRUST AS
             $$
-                Some(::pgx::iter::SetOfIterator::new(names.into_iter().map(|maybe| maybe.map(|name| name.to_string() + " was booped!"))))
+                Ok(::pgx::iter::SetOfIterator::new(names.into_iter().map(|maybe| maybe.map(|name| name.to_string() + " was booped!"))))
```
workingjubilee, line 265: Needs to double-wrap.
```suggestion
                Ok(Some(::pgx::iter::SetOfIterator::new(names.into_iter().map(|maybe| maybe.map(|name| name.to_string() + " was booped!")))))
```

workingjubilee (commented):
Nice!

plrust/src/tests.rs
eeeebbbbrrrr: Yeah, this one just popped up in CI.  :face_palm:

plrust/src/user_crate/mod.rs
```diff
@@ -428,8 +428,8 @@ mod tests {
             let generated_lib_rs = generated.lib_rs()?;
             let imports = crate::user_crate::crating::shared_imports();
             let bare_fn: syn::ItemFn = syn::parse2(quote! {
-                fn #symbol_ident(arg0: &str) -> Option<String> {
-                    Some(arg0.to_string())
+                fn #symbol_ident(arg0: &str) -> std::result::Result<Option<String>, Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
eeeebbbbrrrr: Will do!

plrust/src/tests.rs
eeeebbbbrrrr: This kinda thing makes it look like I didn't run the tests locally first.  EVEN THO I DID I SWEAR

eeeebbbbrrrr:
This PR now requires that https://github.com/tcdi/pgx/pull/1007 be merged and then we release a pgx v0.7.0-beta.1.  and then update plrust to use that.  then we can merge this.

eeeebbbbrrrr:
This is ready to merge, but CI is going to fail b/c [c41a7e3](https://github.com/tcdi/plrust/pull/158/commits/c41a7e34fb82063382ff1b35062fe40616c5c9fa) isn't in the `main` branch yet, and the user function Cargo.toml gets `trusted-pgx` from git's `main` branch.

plrust/src/tests.rs
eeeebbbbrrrr: This is all been fixed up in pgx 0.7.0-beta.1

PR #161, opened 2023-01-14
On Linux, `dlopen()` user functions from a memory-mapped file 

eeeebbbbrrrr:
When running on Linux we create an anonymous file (via `memfd_create()`), write the user function shared library bytes to it, and "dlopen()" from there.

On all other platforms we maintain our current solution of writing a temporary file to the configured `plrust.work_dir`.

### DIFF
--- a/plrust/Cargo.toml
+++ b/plrust/Cargo.toml
@@ -28,6 +28,7 @@ target_postgrestd = []
 cfg-if = "1" # platform conditional helper
 once_cell = "1.7.2" # polyfills a nightly feature
 semver = "1.0.14"
+memfd = "0.6.2" # for loading user function .so
 
 # pgx core details
 pgx = { version = "0.7.0-beta.0" }
--- a/plrust/Cargo.toml
+++ b/plrust/Cargo.toml
@@ -28,7 +28,6 @@ target_postgrestd = []
 cfg-if = "1" # platform conditional helper
 once_cell = "1.7.2" # polyfills a nightly feature
 semver = "1.0.14"
-memfd = "0.6.2" # for loading user function .so
 
 # pgx core details
 pgx = { version = "0.7.0-beta.0" }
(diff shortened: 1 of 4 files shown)


workingjubilee:
Rust makes these things easy but these things are still hard, so this amazingly only has _two_ failing tests.

eeeebbbbrrrr:
> Rust makes these things easy but these things are still hard, so this amazingly only has _two_ failing tests.

haha.  Trying to figure them out now.  Bizarre

plrust/Cargo.toml
```diff
@@ -28,6 +28,7 @@ target_postgrestd = []
 cfg-if = "1" # platform conditional helper
 once_cell = "1.7.2" # polyfills a nightly feature
 semver = "1.0.14"
+memfd = "0.6.2" # for loading user function .so
```
thomcc, line 31: Does this need to be inside `[target.'cfg(target_os="linux")'.dependencies]`?

plrust/src/user_crate/ready.rs
```diff
@@ -27,13 +37,60 @@ impl FnReady {
...
+            mfd.add_seal(memfd::FileSeal::SealSeal)?;
+
+            // and write the shared_object bytes
+            mfd.as_file().write_all(&shared_object)?;
+
+            // generate a direct filename to the underlying raw file descriptor that `mfd` created
+            let raw_fd = mfd.as_raw_fd();
+            let filename = format!("/proc/self/fd/{raw_fd}");
+
+            // finally, load the library.  When this block ends the memory-mapped file will be
```
thomcc, line 69: Nit: memory-mapping is a different thing (you *can* memory-map anonymous files made with `memfd_create`, but they aren't always memory mapped).
```suggestion
            // finally, load the library.  When this block ends the anonymous file will be
```

thomcc (approved):
Largely looks good. Does the `memfd` crate compile on all non-linux targets? If not we should put it behind `[target.'cfg(target_os="linux")'.dependencies]`

plrust/Cargo.toml
eeeebbbbrrrr: It doesn’t need to but I suppose it should. Will fix

PR #162, opened 2023-01-15
Hack for multi-target postgrestd

thomcc:
To use this make sure that if you're on x86_64, then the environment is like
- `CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc`
- `CARGO_TARGET_AARCH64_POSTGRES_LINUX_GNU_LINKER=<same thing>`
And vice versa for x86_64. Then it should work. Depends on https://github.com/tcdi/postgrestd/pull/18


eeeebbbbrrrr:
Thanks for the PR.  I'm taking this over as part of an upcoming PR.

PR #164, opened 2023-01-15
upgrade pgx dependency to 0.7.0-beta.1

eeeebbbbrrrr:
This also changes the expected branch for the "trusted-pgx" dependency to the `main` branch.

~~Because of that change, CI is going to fail.  But I'll merge anyways and then it should pass in the main branch.  Tests pass locally for me.~~

No, it shouldn't.  That crate is already in `main`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -928,9 +928,9 @@ dependencies = [
 
 [[package]]
 name = "pgx"
-version = "0.7.0-beta.0"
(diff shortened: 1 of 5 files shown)


eeeebbbbrrrr:
#158 is waiting on this (which I'll rebase and fixup later tonight) and the additional work for cross-compile is waiting on this too.

PR #165, opened 2023-01-16
More Cross Compilation Work

eeeebbbbrrrr:
I'll come back to this tomorrow...

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,36 +36,43 @@ Failure to do so will cause unexpected behavior around the DROP FUNCTION, DROP S
 
 Additionally, there are two `postgresql.conf` settings that must be configured:
 
-Option | Description
---------------|-----------
-`plrust.pg_config` | The full path of the `pg_config` binary (required).
-`plrust.work_dir` | The directory where pl/rust will build functions with cargo (required).
-`plrust.tracing_level` | A [tracing directive][docs-rs-tracing-directive]. (Default `info`)
+| Option                  | Type    | Description                                                 | Required | Default  |
+|-------------------------|---------|-------------------------------------------------------------|----------|----------|
+| `plrust.pg_config`      | string  | The full path of the `pg_config` binary                     | yes      | <none>.  |
+| `plrust.work_dir`       | string  | The directory where pl/rust will build functions with cargo | yes      | <none>   |
+| `plrust.tracing_level`  | string  | A [tracing directive][docs-rs-tracing-directive]            | no       | `'info'` |
+| `plrust.use_postgrestd` | boolean | Should PL/Rust use `postgrestd` for user functions?         | no       | `true`   |
 
 [github-pgx]: https://github.com/zombodb/pgx
 [github-fpm]: https://github.com/jordansissel/fpm
 [docs-rs-tracing-directive]: https://docs.rs/tracing-subscriber/0.3.11/tracing_subscriber/filter/struct.EnvFilter.html
 
 # Using PL/Rust with Postgrestd
-PL/Rust currently supports being used with an experimental fork of Rust's std entitled `postgrestd` which supports an
-`x86_64-postgres-linux-gnu` target. This is built via the `build` script, which pulls in `postgrestd`.
+PL/Rust currently supports, and in fact defaults to, being used with a fork of Rust's std entitled `postgrestd` which 
+supports both `x86_64-postgres-linux-gnu` and `aarch64-postgres-linux-gnu` targets.
+
+When using `postgrestd`, PL/Rust disallows the ability for user functions to use the filesystem, execute processes, and
+many other things, to meet Postgres' requirements of a "Trusted Procedural Language".
+
+These Rust targets are built via the [`build`](plrust/build) script, which pulls in `postgrestd` from https://github.com/tcdi/postgrestd.
+
 Doing so places a copy of the necessary libraries used by Rust for `std` into the appropriate "sysroot",
 which is the location that rustc will look for building those libraries.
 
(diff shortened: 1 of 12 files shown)


eeeebbbbrrrr:
I'm going to work on the PR description above and then also work on the changes to README.md to talk about cross-compiling, "trusted", and the `trusted-pgx` crate.  Neither of those should hold up a review and a merge.

I think this is in good (enough) shape and passes my smoke testing locally.

README.md
```diff
@@ -1,73 +1,93 @@
...
 
-Further usage requires setting the `PLRUST_TARGET` to the appropriate tuple, as this is still an experimental feature,
-and so is not currently set by default. Passing `--features target_postgrestd` to `cargo` also works. These are subtly different:
+`postgrestd` and the `trusted` feature flag are **not** supported on other platforms.  As such, PL/Rust cannot be 
+considered fully trusted on those platforms.
 
-- `PLRUST_TARGET` is evaluated when the PL/Rust handler builds a function
-- `--features target_postgrestd` is evaluated when the PL/Rust handler is built
-- `PLRUST_TARGET` prevails over `target_postgrestd` if both are set
+## TODO:  PL/Rust is also Cross Compiler
```
thomcc, line 53:
```suggestion
## TODO:  PL/Rust is also a Cross Compiler
```

plrust/src/error.rs
```diff
@@ -32,6 +30,6 @@ pub(crate) enum PlRustError {
     ParsingCodeBlock(syn::Error),
     #[error("Parsing error at span `{:?}`", .0.span())]
     Parse(#[from] syn::Error),
-    #[error("No plrust.plrust_proc entry for ({0}, {1})")]
-    NoProcEntry(pgx::pg_sys::Oid, CompilationTarget),
+    #[error("plrust.plrust_proc.so bytes are NULL")]
+    NullPlRustProcSharedLibraryBytes,
```
workingjubilee, line 34: Does this mean they are all `\0`s or that it is a pointer to address 0x0?

thomcc (approved):
Looks great! I don't have a ton of context on the plrust crate's internals so it's mostly just signing off that the parts I do understand look right, and they do!

plrust/src/lib.rs
```diff
@@ -10,6 +10,27 @@ Use of this source code is governed by the PostgreSQL license that can be found
...
+        any(target_arch = "x86_64", target_arch = "aarch64")
+    ))
+))]
+compile_error!("This platform does not support the 'trusted' version of plrust");
+
+cfg_if::cfg_if! {
+    if #[cfg(feature = "trusted")] {
+        /// We are a trusted language handler.  This will cause plrust user functions to be compiled
+        /// with `postgrestd`
+        pub(crate) const TRUSTED:bool = true;
```
workingjubilee, line 26: extremely minor and tedious nit, nonetheless,
```suggestion
        pub(crate) const TRUSTED: bool = true;
```

plrust/src/lib.rs
```diff
@@ -10,6 +10,27 @@ Use of this source code is governed by the PostgreSQL license that can be found
...
+    } else {
+        /// We are NOT a trusted language handler.  This will cause plrust user functions to be compiled
+        /// with the standard Rust `std`
+        pub(crate) const TRUSTED:bool = false;
```
workingjubilee, line 30:
```suggestion
        pub(crate) const TRUSTED: bool = false;
```

plrust/src/target.rs
```diff
@@ -104,27 +113,83 @@ impl CompilationTarget {
...
-                        Ok(host::target_tuple().into())
-                    }
-                }
-            }
-            Err(env::VarError::NotUnicode(s)) => Err(TargetErr::InvalidSpec(s)),
+#[derive(Clone, Copy, PartialEq, Hash, Debug)]
+pub(crate) enum CrossCompilationTarget {
+    X86_64 = 1,
+    Aarch64 = 2,
+}
```
workingjubilee, line 120: Why are we assigning values to these variants? We usually do that in enums we want to be FFI compatible, but there is no repr attribute here.

plrust/src/error.rs
eeeebbbbrrrr: It means the datum is an SQL NULL.

plrust/src/target.rs
eeeebbbbrrrr: Oversight from an earlier incarnation. Will fix. Good catch.

plrust/src/error.rs
eeeebbbbrrrr: The column itself is declared NOT NULL so this would only happen if Postgres broke

workingjubilee: Ah okay.
In such a case we should probably simply panic, but
```suggestion
    #[error("plrust.plrust_proc.so is an SQL NULL despite being NOT NULL")]
    NullPlRustProcSharedLibraryBytes,
```

plrust/src/target.rs
workingjubilee: Yeah, I'm not so concerned about doing so as much as I was just checking that it wasn't being used in an odd way elsewhere that escaped my notice.

workingjubilee (approved):
This looks correct. Thank you.

eeeebbbbrrrr:
> > test tests::tests::pg_plrust_one_sleepy_boi ... FAILED
> 
> Between all the PRs being merged, the assert somehow made its return...?

It seems to be intermittent. Which is another level of awesome.

workingjubilee:
I'm going to merge this over the error and address the issue in followup if needs be.

PR #167, opened 2023-01-17
Rip out sleepy_boi test

workingjubilee:
This test was useful for noticing failures due to problems with PGX's unwinding correctness. It is no longer useful, and has begun to be a source of transient problems.

It has however revealed something else that might be useful to study more closely, but that is something I will do in a followup.


eeeebbbbrrrr (approved):
Zzzzz

PR #168, opened 2023-01-17
version bump and housekeeping

eeeebbbbrrrr:
- update crate versions to 0.7.0-beta.1 to match the pgx version we use
- remove nix/flake stuff
- update copyright headers
- rewrite "ZomboDB" references to "tcdi", except in some early source files


workingjubilee (approved):
Do we really need license headers in every file? They can't be compiled separately of the rest of the text.

eeeebbbbrrrr:
> Do we really need license headers in every file? They can't be compiled separately of the rest of the text.

I don’t even know. Better safe than sorry even tho it probably doesn’t matter.

PR #173, opened 2023-01-23
get rid of our custom `plrust.plrust_proc` 

eeeebbbbrrrr:
... and instead create a custom json structure to be used in `pg_catalog.pg_proc.prosrc`, using serde.

This also means we don't have to hook "DROP FUNCTION" and "DROP SCHEMA" statements, so all that code is deleted.

Do some other drive-by cleanups along the way w.r.t error handling.

/cc @JohnHVancouver 

One possible TODO would be to create a UDF like `plrust.show_function(regproc)` as a (less than optimal) replacement for psql's `\sf func_name`, which now shows a json object.

Another possible TODO would be to encode the shared library bytes as base64.  Right now we just use serde's `Vec<u8>` serialization format, which is probably quite terrible.

Any thoughts?


plrust/src/prosrc.rs
```diff
@@ -0,0 +1,135 @@
...
+use crate::error::PlRustError;
+use crate::pgproc::PgProc;
+use crate::target;
+use crate::target::CompilationTarget;
+use crate::user_crate::{FnReady, UserCrate};
+
+#[derive(Default, Debug, Serialize, Deserialize)]
+struct ProSrcEntry {
+    src: String,
+    lib: BTreeMap<String, Vec<u8>>,
```
workingjubilee, line 27: Wait. What do each of these Strings represent? Isn't one a compile target? Why not just use that type?

plrust/src/plrust.rs
```diff
@@ -144,9 +141,8 @@ pub(crate) fn crate_name(db_oid: pg_sys::Oid, fn_oid: pg_sys::Oid) -> String {
     // NB:  This once included the compiling host's target triple as part of the crate name for
     // reasons about restoring a database to the same platform.
     //
-    // This isn't necessary as our plrust.plrust_proc "catalog" table tracks the .so binaries per
-    // target triple, so if we are restored to a different platform then the .so binary for this
-    // platform won't be used.
+    // This isn't necessary a we store the .so binaries by target triple in `pg_catalog.pg_proc.prosrc`,
```
workingjubilee, line 144:
```suggestion
    // This isn't necessary as we store the .so binaries by target triple in `pg_catalog.pg_proc.prosrc`,
```

plrust/src/prosrc.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    // fabricate a FnLoad version of the UserCrate so that we can "load()" it -- tho we're
+    // long since past the idea of crates, but whatev, I just work here
+    let built = UserCrate::built(
+        pg_proc.xmin(),
+        db_oid,
+        pg_proc_oid,
+        this_target.clone(),
+        so_bytes,
+    );
```
workingjubilee, line 130: Making a note here to myself of something we probably want to change if bypassing the FSM is being done, but not currently aware of how else it should be.

workingjubilee (commented):
Apparently, serde_json will simply map a `Vec<u8>` into a set of `Number`s: https://github.com/serde-rs/json/blob/993e7a6eeaa89b39329d245e63879d7913ed1a41/src/value/ser.rs#L147-L150

plrust/src/prosrc.rs
eeeebbbbrrrr: Oh. Sure. Duh. Can make CompilationTarget derive the serde traits too. Will fix

eeeebbbbrrrr:
> Apparently, serde_json will simply map a `Vec<u8>` into a set of `Number`s: https://github.com/serde-rs/json/blob/993e7a6eeaa89b39329d245e63879d7913ed1a41/src/value/ser.rs#L147-L150

Yup. 

Is that good enough or should we do something like base64 or encode it as Postgres would encode a bytea type as text or something else?  

It’s not super performance critical as it only happens the first time we load a version of the function but we don’t need to introduce a lot of copying or CPU overhead either.

workingjubilee:
We should definitely base64 encode it or apply compression or really anything, because e.g. if the resulting binary contains any string data, then for 3 ASCII characters, serde will serialize them as something like `[66,105,103]`, which is 2.6 times as big as `"Big"`, and then about another 3 or 4 per additional character.

eeeebbbbrrrr:
> Whatever we do, it would be a good idea to do something... prefix it with a magic number, add a column or json field marking a version, whatever... that describes the encoding so that we can upgrade the serialization/encoding ever, though given most deployments are still prototypes, I think leaving that as a 1.0 todo is fine.

I've pushed some changes to introduce a `SharedLibrary` type that has an `encoding` enum member and an `encoded` String member.  This'll let us change our minds later while still being compatible with anything that might already exist in the wild.

And now we "`gzip -9`" the .so bytes, then base64 those.  It's nearly 10x smaller than the serde version.

JohnHVancouver:
> plrust.show_function(regproc) as a (less than optimal) replacement for psql's \sf func_name

I wonder if it's useful to provide a view of pg_proc as well where the prosrc is handled properly? Or do you think it'd be good enough to document these and have examples on how you would query it so it showed what it did previously.

> And now we "gzip -9" the .so bytes, then base64 those. It's nearly 10x smaller than the serde version.

Nice! IIUC it'll just be a one-time transaction overhead at worst so it seems fairly benign.

eeeebbbbrrrr:
> Nice! IIUC it'll just be a one-time transaction overhead at worst so it seems fairly benign.

Once per backend connection, even.  So much more coarse than "transaction", assuming long-lived connections.

> I wonder if it's useful to provide a view of pg_proc as well ...

I'm not excited about trying to maintain a view on top of system catalogs.  I think exposing a `plrust.show_function()` ought to be enough.  No matter what we do, it's something the user will need to directly do anyways -- it's not like we can also replace psql's `\sf` command.

workingjubilee:
Excellent!

Yeah, I'm not too worried about the performance overhead here, and we can dial in the amount of unzip/zip we do later, just about not having these things use up a ton of space in the serialized form... base64 adding 1/3rd as many bytes is a far cry from serde adding as much as 4 bytes per byte, and on top of function fatness from LLVM inlining (which is usually a good thing for perf, but can put a hard hit on binary size), an idiomatic Rust function has a few kilobytes of error messages or something like that. :wink:

workingjubilee (approved):
So just to make sure I understand: because these system catalog tables are automatically maintained by Postgres in a correct fashion, we don't really have to worry about them falling out of date, and that's why we can avoid hooking into CREATE OR REPLACE FUNCTION and so on? Sounds good to me.

eeeebbbbrrrr:
> So just to make sure I understand: because these system catalog tables are automatically maintained by Postgres in a correct fashion, we don't really have to worry about them falling out of date, and that's why we can avoid hooking into CREATE OR REPLACE FUNCTION and so on? Sounds good to me.

Right, by the time Postgres calls our 

```rust
#[pg_extern]
unsafe fn plrust_validator(fn_oid: pg_sys::Oid, fcinfo: pg_sys::FunctionCallInfo) { ... }
```

function, the row in `pg_catalog.pg_proc` already exists (it has to so that Postgres can hold an exclusive lock on it so concurrent backends can't also create a same-named function).  

So we just change the value of its `prosrc` column to be our fancy json structure and write out a new version.  It's fine for us to do that because, again, the current transaction already has a lock on the row.  

In the "CREATE" case the initial value of the `prosrc` column is just whatever the user put in the `AS $$ ... $$` block.  So we do our compilation steps and rewrite the row, changing only the `prosrc` column to our format.

In the "CREATE OR REPLACE" case the initial value is our json structure containing all the things from the previous incarnation of the row.  So in that case we deserialize it, pop in the new compiled bytes in the map and rewrite the row.

Our distinction between the two is only if the `prosrc` value we have deserializes into our `ProSrcEntry` structure.  But even if the user managed to give us an `AS $$...$$` block that is a properly formed `ProSrcEntry` in json, we won't use any of it -- we replace what's there with the known-to-be-just-compiled-by-plrust .so bytes.

And we have no concern about "DROP FUNCTION" or "DROP SCHEMA CASCADE" because we no longer have other table rows scattered all over the database.  It's all nice and tidy.

Postgres [docs say](https://www.postgresql.org/docs/current/catalog-pg-proc.html):

```
prosrc text

This tells the function handler how to invoke the function. It might be the actual source code of the function for interpreted languages, a link symbol, a file name, or just about anything else, depending on the implementation language/call convention.
```

The founding fathers probably didn't envision storing a base64-encoded gzipped natively compiled rust shared library, but hey.

plrust/src/prosrc.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    // Update the catalog entry in `pg_catalog.pg_proc.prosrc`
+    unsafe {
+        // SAFETY:  We know that `relation` is valid because we made it above, the `ctid` represents
+        // a valid row on disk because the SysCache gave it to us, and `heap_tuple` is a valid pointer
+        // to a `pg_sys::HeapTupleData` because `.into_trigger_datm()` gave that to us above
+        pg_sys::CatalogTupleUpdate(relation.as_ptr(), &mut ctid, heap_tuple);
+
+        // Advance command counter so new tuple can be seen by validator
+        pg_sys::CommandCounterIncrement();
```
JohnHVancouver, line 147: TIL, nice catch

JohnHVancouver:
> I'm not excited about trying to maintain a view on top of system catalogs. I think exposing a `plrust.show_function()` ought to be enough. No matter what we do, it's something the user will need to directly do anyways -- it's not like we can also replace psql's `\sf` command.

Hmm that's fair. If people really care they can always make their own, need to parse the json for prosrc (`prosrc::json->'src'`) I suppose.

plrust/src/pgproc.rs
```diff
@@ -25,17 +26,37 @@ impl Drop for PgProc {
...
+
+    pub(crate) fn relation() -> PgRelation {
+        unsafe {
+            // SAFETY:  [`pg_sys::ProcedureRelationId`] is a compiled-in relation oid, and
+            // [`pg_sys::AccessShareLock`] is a valid lock value
+            PgRelation::with_lock(pg_sys::ProcedureRelationId, pg_sys::AccessShareLock as _)
+        }
+    }
+
+    /// Return a copy of the backing [`pg_sys::HeapTupleData`] allocated in the `CurrentMemoryContext`
```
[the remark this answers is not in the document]
JohnHVancouver: Oh right that's obvious

PR #175, opened 2023-01-24
Only specify CARGO_TARGET_..._LINKER if not already specified, and ensure it and `STD_TARGETS` are exported and avaliable to `./run`.

thomcc:
Reported on discord.

### DIFF
--- a/plrust/build
+++ b/plrust/build
@@ -3,17 +3,21 @@ set -xe
 
 if [ -z "$STD_TARGETS" ]; then
     # if none specified, always build for these two targets
-    STD_TARGETS="x86_64-postgres-linux-gnu aarch64-postgres-linux-gnu"
+    export STD_TARGETS="x86_64-postgres-linux-gnu aarch64-postgres-linux-gnu"
 fi
 
 # and depending on the platform we're building on, we need to set a linker flag for the other
 # this'll get hairy when we support more platforms and we should port this script to Rust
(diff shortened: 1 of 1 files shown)


plrust/build
```diff
@@ -3,17 +3,21 @@ set -xe
 
 if [ -z "$STD_TARGETS" ]; then
     # if none specified, always build for these two targets
-    STD_TARGETS="x86_64-postgres-linux-gnu aarch64-postgres-linux-gnu"
+    export STD_TARGETS="x86_64-postgres-linux-gnu aarch64-postgres-linux-gnu"
```
workingjubilee, line 6: This should instead be explicitly passed to `./run install`, to indicate the dependency.

workingjubilee: This was how it historically was, but it was undone as part of #165 changes.

thomcc: It needs to be passed to clean too, now that cross compilation exists.

plrust/build
```diff
@@ -46,8 +46,8 @@ fi
         cd ./postgrestd
     fi
     rm --force rust-toolchain.toml
-    ./run clean
-    ./run install
+    STD_TARGETS="$STD_TARGETS" ./run clean
```
workingjubilee, line 49: I'm not sure why this acquired a `./run clean` given that `./run install` already does what should be enough but I am gonna figure that out later.

workingjubilee (approved):
Nice, should be good!

PR #177, opened 2023-01-25
Test `#[export_name]` is blocked

workingjubilee:
This adds a new test and rearranges the unsafe test functions slightly (which has a somewhat confusing diff because the tests are very similar).


eeeebbbbrrrr (approved):
LGTM

PR #181, opened 2023-01-25
Remove the `plrust.pg_config` GUC

eeeebbbbrrrr:
Turns out Postgres itself, via its internal `get_configdata()` function knows everything `pg_config` does.  So we convert all the `ConfigData` key/value pairs into environment variables for pgx-pg-sys to use during its build.rs.  

This relies on the changes in https://github.com/tcdi/pgx/pull/1019 and https://github.com/tcdi/pgx/pull/1020.

### DIFF
--- a/plrust/src/user_crate/build.rs
+++ b/plrust/src/user_crate/build.rs
@@ -111,7 +111,7 @@ impl FnBuild {
         //
         // This turned out to be an unwanted bit of user, system, and operational complexity.
         //
-        // Instead, we tell tell the environment that "pg_config" is described as environment
+        // Instead, we tell the environment that "pg_config" is described as environment
         // variables, and set every property Postgres can tell us (which is essentially how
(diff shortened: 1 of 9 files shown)


plrust/src/user_crate/build.rs
```diff
@@ -192,3 +207,39 @@ impl FnBuild {
...
+                // done so by the `get_configdata()` call above
+                let name = CStr::from_ptr(e.name);
+                let setting = CStr::from_ptr(e.setting);
+                values.push((
+                    name.to_string_lossy().to_string(),
+                    setting.to_string_lossy().to_string(),
+                ))
+            }
+
+            values.into_iter()
```
JohnHVancouver, line 242: IIUC this memory context will be destroyed when `values.into_iter()` falls out of scope?

JohnHVancouver (approved):

workingjubilee (approved):
Less configuration is an improvement!

PR #184, opened 2023-01-25
Cleanup source crate after compilation fails

JohnHVancouver:
Not all users of PL/Rust will have access to the file-system to manually clean up source code when compilation fails.

Add cleanup when compilation fails, the error is already outputted to the end user.

Also fix compilation on x86 mac, it complains about crate_name being behind a shared reference otherwise

### DIFF
--- a/plrust/src/user_crate/build.rs
+++ b/plrust/src/user_crate/build.rs
@@ -146,7 +146,7 @@ impl FnBuild {
                 feature = "force_enable_x86_64_darwin_generations"
             ))]
             let crate_name = {
-                let mut crate_name = crate_name;
+                let mut crate_name = crate_name.clone();
                 let next = crate::generation::next_generation(&crate_name, true)
                     .map(|gen_num| gen_num)
                     .unwrap_or_default();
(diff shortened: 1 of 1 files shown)


JohnHVancouver:
This doesn't help if Postgres was `pkill -9` or crashes during compilation. I wonder if it's worth while adding a check to clean-up the `work_dir` during start in `_PG_init`. Something like clean up every file that's `work_dir/plrust*`

plrust/src/user_crate/build.rs
```diff
@@ -180,15 +180,23 @@ impl FnBuild {
...
                         .expect("Reading generated `lib.rs` to output during error")
                         .header("Source Code:")
-                }))?
+                }));
+
+            // Clean up on error
+            std::fs::remove_dir_all(&self.crate_dir).wrap_err(format!(
+                "Problem deleting temporary crate directory at '{}'",
+                self.crate_dir.display()
+            ))?;
```
workingjubilee, line 197: In the error case, if we fail to clean up, we should not fail to propagate the actual error message. This would do so. Attempt the cleanup without doing error propagation.

plrust/src/user_crate/build.rs
```diff
@@ -146,7 +146,7 @@ impl FnBuild {
                 feature = "force_enable_x86_64_darwin_generations"
             ))]
             let crate_name = {
-                let mut crate_name = crate_name;
+                let mut crate_name = crate_name.clone();
```
workingjubilee, line 149: Thank you!

workingjubilee (changes requested):

plrust/src/user_crate/build.rs
JohnHVancouver: Ahh yea that makes sense. I wasn't sure if it would happen to be problematic but it makes more sense to always throw the relevant compilation error

workingjubilee (approved):

Issue #183, opened 2023-01-25
plrust can cause ERRORs in databases without the extension


eeeebbbbrrrr:
plrust hooks the "process utility statement" and the extension library is loaded by the postmaster as a `shared_preload_libraries` entry.  This means every backend connection to every database actually has plrust code running.  Even those databases without the plrust extension installed.

Unfortunately, plrust does something in that hook where during an ALTER FUNCTION statement, it'll raise an ERROR if plrust isn't installed in the database.

This is easy enough to fix, and I will tomorrow.

eeeebbbbrrrr:
Fixed in #185

PR #185, opened 2023-01-26
fix issue #183

eeeebbbbrrrr:
`plrust` will no longer raise an ERROR if a function is being ALTERed in a database that doesn't also have the `plrust` extension created.


workingjubilee (approved):
Nice! I have almost no idea how this works but it looks good to me.

eeeebbbbrrrr:
> Nice! I have almost no idea how this works but it looks good to me.

It doesn't

PR #188, opened 2023-01-27
Ignore Err if remove_dir_all fails

workingjubilee:


plrust/src/user_crate/build.rs
```diff
@@ -191,10 +191,7 @@ impl FnBuild {
                 }));
 
             // Clean up on error but propagate the more relevant error
```
thomcc, line 193: This comment is wrong now.

plrust/src/user_crate/build.rs
```diff
@@ -191,10 +191,7 @@ impl FnBuild {
...
-            std::fs::remove_dir_all(&self.crate_dir).wrap_err(format!(
-                "Problem deleting temporary crate directory at '{}'",
-                self.crate_dir.display()
-            ));
+            let _ = std::fs::remove_dir_all(&self.crate_dir);
```
thomcc, line 194: Maybe this should log a warning about the failure?

plrust/src/user_crate/build.rs
thomcc: Never mind, I see, it's not.

thomcc (approved):

plrust/src/user_crate/build.rs
```diff
@@ -190,11 +190,10 @@ impl FnBuild {
...
                 }));
 
-            // Clean up on error but propagate the more relevant error
-            std::fs::remove_dir_all(&self.crate_dir).wrap_err(format!(
-                "Problem deleting temporary crate directory at '{}'",
-                self.crate_dir.display()
-            ));
+            // Clean up on error but don't let this error replace our user's error!
+            if let Err(e)= std::fs::remove_dir_all(&self.crate_dir) {
+                pgx::warning!("Problem during removing crate directory: {e}")
```
eeeebbbbrrrr, line 195: I think this ought to go to the LOG instead:  `pgx::log!()`.

My primary concern is that since plrust is basically ignoring the error, there's no need to make it such that the user will see it in default Postgres installations.  We're also relying on `std::fs::remove_dir_all()` to not leak any "sensitive" information in its error that might give away filesystem layout or something.

eeeebbbbrrrr (changes requested):

plrust/src/user_crate/build.rs
workingjubilee: `io::Error` does not contain any such directory path messaging that I am aware of, it will only disclose the nature of the error that may be relevant to the user (e.g. "out of space on device"). If this is still a concern, I will simply remove the message.

eeeebbbbrrrr: I know that it doesn't now, and @thomcc just said that since he's the author of that type it likely never will.  But...

Really, the concern is that defaulting to a level that the user is highly likely to see, for a thing we're otherwise ignoring and they likely can't do anything about, just isn't valuable information for them.  Postgres doesn't do this in most cases -- it's rare for it to spit out a WARN message.

workingjubilee: If the deployment is on a managed database service, and thus incapable of directly configuring the instance, they may still use the user console provided by the managed database service to alter the backing block storage (in the simple case: by upgrading it).

eeeebbbbrrrr: It's much more common for the user running CREATE FUNCTION to **not** be the DBA/sysadmin.  

If the intent is that we require some kind of action to be taken because we couldn't delete a directory, then we ought to just raise it as an ERROR.  Otherwise, it's pretty useless information for the user, but definitely valuable for the LOG.

workingjubilee: That is not my experience, as I try to respect the principle of least privilege after setting up a database. That means I would tend to log in as a user even when I have superuser access. I would be the only user, yet I have two or even more logins to the same DB. It's also common nowadays to not have a strict divide between developers and DBAs.

eeeebbbbrrrr (approved):

PR #192, opened 2023-01-30
Add a test for include_str!

### LINKED ISSUE #172: Need to disallow Rust's `include_xxx!()` macros
This plrust function should not compile, let alone work:

```postgres
create function include_str() returns text language plrust as 
$$ 
   let s = include_str!("/etc/passwd"); 
   Ok(Some(s.into())) 
$$;
```

@workingjubilee is this something we can block in `postgrestd`?  It appears to be in `std`:  https://doc.rust-lang.org/std/macro.include_str.html

thomcc:
~~CI intentionally fails here, it requires https://github.com/tcdi/postgrestd/pull/23 for it to pass.~~ That landed, so it passes now.

Fixes #172

### DIFF
--- a/plrust/src/tests.rs
+++ b/plrust/src/tests.rs
@@ -450,6 +450,26 @@ mod tests {
         Ok(())
     }
 
+    #[cfg(feature = "trusted")]
+    #[pg_test]
+    #[search_path(@extschema@)]
+    #[should_panic]
+    fn postgrestd_no_include_str() -> spi::Result<()> {
+        let definition = r#"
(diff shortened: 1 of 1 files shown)


thomcc:
Evidence that CI passes with my postgrestd patches: https://github.com/tcdi/plrust/actions/runs/4046720922

workingjubilee (approved):
Seems good.

eeeebbbbrrrr (approved):
LGTM!  Thanks @thomcc

PR #193, opened 2023-01-30
Set a `$PATH` when there isn't one

eeeebbbbrrrr:
Debian-based distro-managed Postgres installations (such as Ubuntu) seem to clear the `$PATH` environment variable when Postgres starts.  This means that plrust can't find the system commands such as `cargo` and `cc` that it needs to compile user functions (there are probably others).

If the `$PATH` is missing (or the empty string), we'll hardcode it to `~/.cargo/bin:/usr/bin` when executing `cargo`.  However, if the user has set the `plrust.PATH_override` GUC, then we will **always** use that when executing `cargo`.

### DIFF
--- a/plrust/src/user_crate/mod.rs
+++ b/plrust/src/user_crate/mod.rs
@@ -252,22 +252,28 @@ pub(crate) fn oid_to_syn_type(type_oid: &PgOid, owned: bool) -> Result<syn::Type
 }
 
 /// Builds a `Command::new("cargo")` with necessary environment variables pre-configured
-pub(crate) fn cargo() -> Command {
+pub(crate) fn cargo() -> eyre::Result<Command> {
     let mut command = Command::new("cargo");
 
     if let Some(path) = PLRUST_PATH_OVERRIDE.get() {
         // we were configured with an explicit $PATH to use
         command.env("PATH", path);
     } else {
-        let path = std::env::var("PATH");
+        let is_empty = match std::env::var("PATH") {
+            Ok(s) if s.trim().is_empty() => true,
+            Ok(_) => false,
+            Err(VarError::NotPresent) => true,
+            Err(e) => return Err(eyre::eyre!(e)),
+        };
 
-        if path == Err(VarError::NotPresent) || path == Ok(String::default()) {
+        if is_empty {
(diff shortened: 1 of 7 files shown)


plrust/src/user_crate/mod.rs
```diff
@@ -248,6 +251,35 @@ pub(crate) fn oid_to_syn_type(type_oid: &PgOid, owned: bool) -> Result<syn::Type
...
+    } else {
+        let path = std::env::var("PATH");
+
+        if path == Err(VarError::NotPresent) || path == Ok(String::default()) {
+            // the environment has no $PATH, so lets try and make a good one based on where
+            // we'd expect 'cargo' to be installed
+            if let Some(home_directory) = dirs::home_dir() {
+                command.env(
+                    "PATH",
+                    format!("{}/.cargo/bin:/usr/bin", home_directory.display()),
```
thomcc, line 270: This is wrong on windows (I don't *think* it's wrong anywhere else, but am not certain), since `dirs::home_dir()` isn't where cargo places files. The right way would be to use <https://crates.io/crates/home>.

Also, this could use https://doc.rust-lang.org/nightly/std/env/fn.join_paths.html, but that's not that important.

thomcc (approved):
In general this is wrong in several ways on windows (wrong method of locating the path, and using `cmd.env("PATH", ...)` to locate the process doesn't work there, since the path isn't modified until after the process is spawned). But we don't support Windows, so it's fine.

I think it should be fine elsewhere, although perhaps we should use the `home` crate instead of `dirs` here. to locate the cargo home directory.

plrust/src/user_crate/mod.rs
eeeebbbbrrrr: oh, I didn't know about the `home` crate.  I'll switch over to that.

I don't particularly care about Windows support since we don't, but best to future proof the easy things we can, I suppose.

eeeebbbbrrrr:
let me know how you like https://github.com/tcdi/plrust/pull/193/commits/a5e1349cb059b44f44d7e861feb04a5ff39d07f5

thomcc (approved):
Looks great

PR #196, opened 2023-02-03
Remove MacOS-specific "generation"

eeeebbbbrrrr:
This is an initial stab at removing the MacOS-specific "generation" stuff and instead using the `pg_proc` entry's "xmin" and "cmin" values to formulate one that's based on the tuple's immutable characteristics.  And this does it for all platforms.

We also now make a distinction between the "crate_name" which includes this number and the "symbol name" which does not.  We store the symbol name as an `Option<String>` in our json value in `prosrc` and will use a backwards compatible default for Linux when loading a user function that doesn't have the `"symbol"` property.  No concern was given to being backwards compatible for any existing MacOS plrust functions because... MacOS.


plrust/src/pgproc.rs
```diff
@@ -76,6 +76,28 @@ impl PgProc {
...
+                .unwrap_unchecked() // SAFETY: t_data will never be null and `cmin()` is called in a potentially hot path
+                .t_choice
+                .t_heap
+                .t_field3
+                .t_cid
+        }
+    }
+
+    #[inline]
+    pub(crate) fn generation_number(&self) -> u64 {
```
eeeebbbbrrrr, line 97: Couldn't decide if we wanted a stronger type than `u64`.  I considered `struct GenerationNumber { xmin: TransactionId, cmin: CommandId }` but didn't want to do all that typing if this idea turns out bad.

workingjubilee: A u64 is fine for an incrementing counter, we can lock things down later if needs be.

plrust/src/plrust.rs
```diff
@@ -136,7 +136,15 @@ pub(crate) fn compile_function(fn_oid: pg_sys::Oid) -> eyre::Result<Output> {
     Ok(this_output.unwrap())
 }
 
-pub(crate) fn crate_name(db_oid: pg_sys::Oid, fn_oid: pg_sys::Oid) -> String {
+pub(crate) fn symbol_name(db_oid: pg_sys::Oid, fn_oid: pg_sys::Oid) -> String {
+    format!("plrust_fn_oid_{}_{}", db_oid.as_u32(), fn_oid.as_u32())
+}
+
+pub(crate) fn crate_name(
```
workingjubilee, line 143: Maybe these should explain when to use the different names?

workingjubilee (approved):
This looks good to me!

plrust/src/plrust.rs
eeeebbbbrrrr: sure, I'll take care of that.

PR #199, opened 2023-02-08
document environment variables...

eeeebbbbrrrr:
...  and remove a set of envvars that we know aren't necessary and might could cause problems.

I think this is all of them.  I went through pgx-pg-sys/build.rs and all of plrust.


thomcc (approved):
Looks good to me!

PR #201, opened 2023-02-13
Test postgrestd/1.67.1

workingjubilee:


thomcc:
I'm assuming we're doing this (and adjusting my code accordingly) but am not going to click merge just in case.

eeeebbbbrrrr:
I don't think this is necessary now that #198 was merged?

thomcc:
Ah, I should have merged this first probably (although that would have caused issues for you, since then I'd need to rebase). Sorry for the hassle my branch caused.

workingjubilee:
It's nbd, honestly.

Issue #200, opened 2023-02-11
Unable to create SRF function with SETOF


rustprooflabs:
I have not been able to create a set returning function using `SETOF`.  I get the same basic error using both a custom type for the return as well as the generic `record` type.

## Use custom type

```sql
CREATE TYPE srf_row AS (id BIGINT);
```

```sql
CREATE FUNCTION spi_srf()
    RETURNS SETOF srf_row
    LANGUAGE plrust
AS
$$
    let query = "SELECT id::BIGINT FROM generate_series(1, 3) id;";

    Spi::connect(|client| {
        let mut results = Vec::new();
        let mut tup_table = client.select(query, None, None)?;

        while let Some(row) = tup_table.next() {
            let id = row["id"].value::<i64>();
            results.push((id,));
        }
        Ok(TableIterator::new(results.into_iter()))
    })

$$
;
```

This fails with the `\errverbose` output:

```
ERROR:  XX000: 
   0: Oid `oid=#83781` was not mappable to a Rust type

Location:
   plrust/src/user_crate/crate_variant.rs:70

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
LOCATION:  lib.rs:191
```

Confirmed the custom type is the `oid` mentioned.

```sql
select oid, typname from pg_type WHERE oid = 83781;
```

```
┌───────┬─────────┐
│  oid  │ typname │
╞═══════╪═════════╡
│ 83781 │ srf_row │
└───────┴─────────┘
```

## Record type

Attempting to return `SETOF record` also fails with the same basic error.

```sql
CREATE FUNCTION spi_srf()
    RETURNS SETOF record
    LANGUAGE plrust
AS
$$
    let query = "SELECT id::BIGINT FROM generate_series(1, 3) id;";

    Spi::connect(|client| {
        let mut results = Vec::new();
        let mut tup_table = client.select(query, None, None)?;

        while let Some(row) = tup_table.next() {
            let id = row["id"].value::<i64>();
            results.push((id,));
        }
        Ok(TableIterator::new(results.into_iter()))
    })

$$
;
```

Same error location on the `RECORDOID`.

```
ERROR:  XX000: 
   0: Oid `oid={#2249, builtin: RECORDOID}` was not mappable to a Rust type

Location:
   plrust/src/user_crate/crate_variant.rs:70

[...]

PR #202, opened 2023-02-13
Fix issue #200

eeeebbbbrrrr:
This adds a `<'a>` lifetime to all user functions so that the few cases where arguments are borrowed (`&'a str`) and return types might borrow (`SetOfIterator<'a, ...>)` can do so.

Also blindly adds the `#[allow(unused_lifetime)]` annotation to all instances of user functions so that compilation warnings won't happen in the cases where all args/return don't borrow.

### DIFF
--- a/plrust/src/user_crate/crate_variant.rs
+++ b/plrust/src/user_crate/crate_variant.rs
@@ -69,7 +69,7 @@ impl CrateVariant {
         let return_type: syn::Type = {
             let bare = oid_to_syn_type(&return_oid, true)?;
             match return_set {
-                true => syn::parse2(quote! { ::std::result::Result<Option<::pgx::iter::SetOfIterator<Option<#bare>>>, Box<dyn ::std::error::Error>> })
+                true => syn::parse2(quote! { ::std::result::Result<Option<::pgx::iter::SetOfIterator<'a, Option<#bare>>>, Box<dyn ::std::error::Error>> })
                     .wrap_err("Wrapping return type")?,
                 false => syn::parse2(
                     quote! { ::std::result::Result<Option<#bare>, Box<dyn ::std::error::Error>> },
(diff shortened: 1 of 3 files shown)


workingjubilee (approved):
Hm. Possibly we should just be writing a function signature that lifts the type into a boxed form.

eeeebbbbrrrr:
> Hm. Possibly we should just be writing a function signature that lifts the type into a boxed form.

I don't follow, which means pgx probably won't be able to follow either...

workingjubilee:
In PL/Rust, there's not as much advantage here for using a borrowed form and in the past we've had safety problems from trying to borrow things that should not have been borrowed. For a function that accepts `&str`, we often might as well just return `String` or `Box<str>`.

This would follow the pattern of [ToOwned](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html).

eeeebbbbrrrr:
> In PL/Rust, there's not as much advantage here for using a borrowed form and in the past we've had safety problems from trying to borrow things that should not have been borrowed. For a function that accepts `&str`, we often might as well just return `String` or `Box<str>`.
> 
> This would follow the pattern of [ToOwned](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html).

We do require that all return values are owned.  plrust enforces that.

However, the `SetOfIterator` type needs a lifetime, so that's what this PR is really addressing.  I just went ahead and hit the couple of argument types that might be borrowed from the caller (the &str and the &[u8]).

And in those two cases, I think it is necessary for those to be zero-copy for performance reasons.  We don't have to give them this `'a` lifetime tho, since, again, a PL/Rust function can only return owned types ('cept for SetOfIterator).

workingjubilee:
Right. I was mostly musing aloud as to future directions rather than a blocking concern.

Issue #204, opened 2023-02-14
pg_dump / pg_restore does not work


JohnHVancouver:
When we pg_dump a PL/Rust function the prosrc is a json.
We don't currently parse the json to recompile if a user has created it as one, it just errors out.

I think we need to do something similar to:

https://github.com/tcdi/plrust/blob/8cd37583559449db694b52b442b9ad80fe92a09b/plrust/src/prosrc.rs#L105-L112

and add it here when parsing?

https://github.com/tcdi/plrust/blob/main/plrust/src/user_crate/mod.rs#L295-L300

Check if code_and_deps can be converted into a ProSrcEntry. If so grab the ProSrcEntry.src otherwise assume not json and continue?

------

For reference I created a PL/Rust function and pg_dump the database

```
pg_dump -d plrust -h localhost -f plrust.sql
psql -f plrust.sql -d restore -h localhost
...
psql:plrust.sql:61: ERROR:  
   0: Parsing `[code]` block: expected one of: `for`, parentheses, `fn`, `unsafe`, `extern`, identifier, `::`, `<`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime

Location:
   plrust/src/user_crate/mod.rs:277
```

workingjubilee:
Uhm. Hm. Can you give a sample of the JSON that is expected, here?

workingjubilee:
...Oh wait, of course. It's the one we store!

eeeebbbbrrrr:
Sounds like something is bugged but the intent is that on a restore (create or replace that provides the json) we ignore everything except the “src” property and just do our full compile process.

Issue #206, opened 2023-02-14
Probably want bulk function compilation


workingjubilee:
Considering the `pg_dump` and `pg_restore` use case: By the time we get to large suites of thousands of functions, we're gonna want to have the ability to rebuild everything in one go.

JohnHVancouver:
OOC what would be the use case? To attempt to compile everything with the latest set of lints? I think you could create a plpgsql wrapper around the information in pg_proc technically. There use to be a `recompile_function` in earlier versions that I think the intent was to be used by read replicas backed when things were stored on disk?

workingjubilee:
Yes. Our current assumption is that the contents of the `pg_dump` someone uses `pg_restore` on should not be trusted, and may not match the current environment, nor configuration, nor etc., and we should rerun the compilation process from the top.

PR #212, opened 2023-02-15
fix issue #204: `pg_restore` works

eeeebbbbrrrr:
To properly handle restoration via `pg_restore` we need to at least try and parse the function "code" given to us as if it were a JSON blob that matches our format in `pg_catalog.pg_proc.prosrc`.  If it does match, we pluck out the "src" property.

If it doesn't match then it's probably just a regular CREATE OR REPLACE statement, so we use it as-is.

This logic is encapsulated inside the new `maybe_extract_source_from_json` function.

This fixes issue #204.  /cc @JohnHVancouver

### DIFF
--- a/plrust/src/prosrc.rs
+++ b/plrust/src/prosrc.rs
@@ -7,6 +7,7 @@ Use of this source code is governed by the PostgreSQL license that can be found 
 */
 
 //! Routines for managing the `pg_catalog.pg_proc.prosrc` entry for plrust functions
+use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::io::prelude::*;
 use std::rc::Rc;
@@ -88,6 +89,14 @@ impl TryFrom<&PgProc> for ProSrcEntry {
     }
 }
 
+impl TryFrom<&str> for ProSrcEntry {
+    type Error = serde_json::Error;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        serde_json::from_str::<ProSrcEntry>(value)
+    }
+}
+
 impl Into<String> for ProSrcEntry {
     fn into(self) -> String {
         serde_json::to_string(&self).expect("unable to serialize ProSrcEntry to json")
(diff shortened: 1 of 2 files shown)


workingjubilee (approved):
Thinking out loud about the alternatives: It would be _nice_ for us to not rely on this kind of inference for this case, since this changes what input we accept significantly, perhaps by figuring out a way to sneakily qualify either the `pg_restore` or the non-`pg_restore` case. But... I'm not sure there's a way to actually do that, and it's ultimately okay either way because we don't trust the input here, either, but instead recompile it from the beginning.

eeeebbbbrrrr:
Maybe if we also signed the code or the entire json blob we could blindly trust it?  Seems like a lot of moving parts and complexity tho. 

Postgres does set the “check function bodies” GUC to off during a pg_restore. Maybe that’s a thing we can use here?  I’d need to look it up to see if it’s a superuser only GUC or what. 

That might be the super nice way to handle this. Then we’d only need to do the compile if the new host is a different architecture. And just mix it in to the whole json blob.

workingjubilee:
Yeah, I think maybe relaxing _slightly_ based on the `check_function_bodies` GUC would be a reasonable followup to this and essentially an alternative to https://github.com/tcdi/plrust/issues/206. We'd proooobably still want to do lint checking and such to make sure we aren't violating some other rule.

I don't think code-signing is a good path forward, that requires us to do far too much reasoning about the trustworthiness of our input and adds cryptography-correctness concerns.

eeeebbbbrrrr:
I’ll look into that tomorrow.

workingjubilee:
Sounds good!
Feel free to go ahead with merging this, obvs, speeding up compilation is extra credit.

eeeebbbbrrrr:
On the other hand, do we even trust what pg_restore gives us?  We have no chain of custody on the original dump file.

workingjubilee:
I guess it depends on whether or not we want to allow the DB admin to shoot themselves in the foot.

eeeebbbbrrrr:
So do we just leave as-is?

workingjubilee:
> The point of having a validator is not to let the call handler omit checks, but to notify the user immediately if there are obvious errors in a CREATE FUNCTION command.) While the choice of exactly what to check is mostly left to the discretion of the validator function, note that the core CREATE FUNCTION code only executes SET clauses attached to a function when `check_function_bodies` is on. Therefore, checks whose results might be affected by GUC parameters definitely should be skipped when `check_function_bodies is off`, to avoid false failures when restoring a dump.

This is what [Writing a Procedural Language Handler](https://www.postgresql.org/docs/15/plhandler.html) has to say about it, as of version 15.

eeeebbbbrrrr:
plrust uses no GUCs that could be changed on the function level. 

And we don’t have a facility for compiling a function after the catalog entry exists (ie, in a future transaction). 

So… I think we’re okay.

eeeebbbbrrrr:
Our validation happens at runtime when the function is first loaded.

workingjubilee:
Right.

I think the code as it currently stands after tcdi/plrust@e00d0be is good. We can contemplate this sort of thing at our leisure (...possibly after getting feedback from a DBA or five?).

PR #214, opened 2023-02-15
Lint uses of `include_str`/`include_bytes`

thomcc:
This is probably required to avoid issues in cases where people `include_bytes!` things like keys and certificates.


thomcc:
Surprised I managed to get all the cases. But it turns out we already had diagnostic items for these.

eeeebbbbrrrr:
ha!  this is great!

workingjubilee (changes requested):
I think you need to add a `#[should_panic]` to `postgrestd_no_include_str` or otherwise change the test.

workingjubilee (approved):
Thanks for taking care of that nit, Jubilee! :woman: :1st_place_medal: :woman:

thomcc:
Thanks for taking care of that nit, Jubilee!

PR #215, opened 2023-02-15
Stop using LTO to build function crates

workingjubilee:
This isn't necessary for compilation, causes problems, and massively slows down compilation.
We don't need to be using 1 CGU either.
These changes may penalize performance slightly. However, compilation times have been enough of a recurring concern that we must balance runtime and compilation time as we proceed.


BradyBonnette (approved):
Looks good, learned something in the process too :)

PR #216, opened 2023-02-16
Block function pointers via a lint

thomcc:
This is less fine-grained than I'd like, but to start with should be fine.


LINTS.md
```diff
@@ -76,3 +76,17 @@ For example, this code pattern is blocked:
...
 const SOMETHING: &str = include_str!("/etc/passwd");
 ```
+
+## `plrust_fn_pointers`
+
+Currently, several soundness holes have to do wuth the interaction between function pointers, implied bounds, and nested references. As a stopgap against these, use of function pointer types are currently blocked. This lint will likely be made more precise in the future.
+
+Note that function types (such as the types resulting from closures as required by iterator functions) are still allowed, as these do not have the issues around variance.
+
+For example, the following code patternF is blocked:
```
eeeebbbbrrrr, line 86: Can I get an F in the chat?

eeeebbbbrrrr (approved):

LINTS.md
thomcc: F

eeeebbbbrrrr:
Another winner.  Thanks!

PR #217, opened 2023-02-16
Block async/await via a lint

thomcc:


LINTS.md
```diff
@@ -90,3 +90,20 @@ fn takes_fn_arg(x: fn()) {
     x();
 }
 ```
+
+## `plrust_async`
+
+Currently async/await are forbidden PL/Rust due to unclear interactions around lifetime and soundness constraints. This may be out of an overabundance of caution. Specifically, code like the following will fail to compile:
```
eeeebbbbrrrr, line 96: “… are forbidden **by** …”

eeeebbbbrrrr (approved):
You’re a machine!

PR #220, opened 2023-02-17
Lint against `include!`, `env!`, `option_env!`, and external modules as well

thomcc:
This is tricky because I don't know a way to test this without breaking my setup for local development (E.g. if I add a uitest for it that depends on postgrestd, I'll no longer be able to bless tests on macOS). So I really don't want to do that.

I also don't know of anything this actually prevents. I guess including a sufficiently rust-like file?

Requires https://github.com/tcdi/postgrestd/pull/28, and if that causes a compiler error, then this approach can't be used.


workingjubilee:
Crossposting for reference:
> Since this is a low priority (we don't have a PoC yet to verify this is needed) I am going to first figure out a way to test this to make sure it works (maybe this is the incentive I need to finally add proper CI here instead of just running PL/Rust's test suite) before merging this.

_Originally posted by @workingjubilee in https://github.com/tcdi/postgrestd/issues/28#issuecomment-1435303254_

thomcc:
This is kind of a pile of unrelated lints, but adding each one ends up conflicting so I didn't want to deal with that more than I have to.

Specifically it blocks:
- `include!` (under postgrestd)
- `env!`/`option_env!` (under postgrestd)
- `mod blah;` where the module is not inline (always). This ideally would just be `#[path = "..."] mod` but there's not really a need to only limit it to that (PL/Rust functions are definitely only allowed to be one file), and that is a lot harder to detect reliably.

thomcc:
Oh I needed to merge https://github.com/tcdi/postgrestd/pull/32 first, oops.

thomcc:
Interesting

thomcc:
The failures are from the pgx 0.7.2 update. It seems like we don't properly pin the version.

workingjubilee (approved):
Cool, this stuff works now.

PR #224, opened 2023-02-21
Lint against leaky functions

thomcc:
~~Unsure why Vec::leak isn't detected. Probably have to detect method use separately, ugh.~~ Fixed.


workingjubilee:
Does it work if you also look for `std::vec::Vec::leak`?

thomcc:
@workingjubilee I had to explicitly check method calls.

The defpath is the path where it's defined (regardless of where it's exported from or how it is imported) so alloc is correct.

thomcc:
Note that I detect but don't test for String::leak, since it's unstable. I could remove that too, but it seems likely to stabilize sooner or later.

workingjubilee:
Better to be aware of nightly functions than not, I think.

plrustc/plrustc/src/lints.rs
```diff
@@ -149,12 +150,66 @@ impl EarlyLintPass for PlrustAsync {
...
+                    PLRUST_LEAKY,
+                    "Leaky functions are forbidden in PL/Rust",
+                    |b| b.set_span(expr.span),
+                );
+            }
+        }
+    }
+}
+
+fn does_expr_call_path(cx: &LateContext<'_>, expr: &Expr<'_>, segments: &[&str]) -> bool {
```
thomcc, line 181: This name is still sort of inaccurate. It's more "is it a path expr or method call for the path", but that's what you want most of the time (because we should catch `let x = Vec::leak; x(vec![a, b, c])`)

thomcc: To be clear, that case is caught. I just added a test for it.

workingjubilee: Oh okay!

eeeebbbbrrrr (approved):

PR #226, opened 2023-02-22
Lint against `dyn Fn{,Mut,Once}`

thomcc:
Wasn't able to get any of these to work like function pointers do, but it seems like they have the same soundness holes nonetheless (which makes sense).

### DIFF
--- a/LINTS.md
+++ b/LINTS.md
@@ -79,7 +79,7 @@ const SOMETHING: &str = include_str!("/etc/passwd");
 
 ## `plrust_fn_pointers`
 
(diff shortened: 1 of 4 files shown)


LINTS.md
```diff
@@ -79,7 +79,7 @@ const SOMETHING: &str = include_str!("/etc/passwd");
 
 ## `plrust_fn_pointers`
 
-Currently, several soundness holes have to do wuth the interaction between function pointers, implied bounds, and nested references. As a stopgap against these, use of function pointer types are currently blocked. This lint will likely be made more precise in the future.
+Currently, several soundness holes have to do wuth the interaction between function pointers, implied bounds, and nested references. As a stopgap against these, use of function pointer types and function trait objects are currently blocked. This lint will likely be made more precise in the future.
```
eeeebbbbrrrr, line 82: s/wuth/with/

eeeebbbbrrrr (approved):
one small existing typo

PR #229, opened 2023-02-23
Cleanup/rewrite plrustc main.rs

thomcc:
Outside of work, I happened to write another rustc_driver wrapper. I managed to make it significantly simpler than what we're doing in plrustc, so here's a commit using that approach instead.

### DIFF
--- a/plrustc/plrustc/src/main.rs
+++ b/plrustc/plrustc/src/main.rs
@@ -9,35 +9,20 @@ extern crate rustc_lint;
 extern crate rustc_lint_defs;
 extern crate rustc_session;
 extern crate rustc_span;
-// extern crate rustc_tools_util;
 
+use rustc_driver::Callbacks;
 use rustc_interface::interface;
-// use rustc_session::parse::ParseSess;
-use rustc_span::symbol::Symbol;
-use std::env;
-use std::path::PathBuf;
-use std::process::exit;
+use std::ffi::OsStr;
+use std::path::{Path, PathBuf};
 
 mod lints;
 
-struct PlRustcPassthroughCallbacks;
-
-impl rustc_driver::Callbacks for PlRustcPassthroughCallbacks {}
-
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr (approved):
Definitely looks less complex.  I can't really review this for correctness, so assuming CI passes, go ahead and merge.

plrustc/plrustc/src/main.rs
```diff
@@ -51,66 +36,24 @@ impl rustc_driver::Callbacks for PlrustcCallbacks {
 }
 
 fn main() {
-    // Snapshot the environment before we call functions which may mess with it.
-    // let env_snapshot = env::vars_os().collect::<Vec<_>>();
     rustc_driver::install_ice_hook();
```
workingjubilee, line 39: :ice_cube:

plrustc/plrustc/src/main.rs
```diff
@@ -127,124 +70,101 @@ fn main() {
...
-    // Invoke compiler, and handle return code.
-    let exit_code = rustc_driver::catch_with_exit_code(move || {
+fn run_compiler<CB: Callbacks + Send>(mut args: Vec<String>, callbacks: &mut CB) -> ! {
+    args.splice(1..1, std::iter::once("--cfg=plrustc".to_string()));
+    std::process::exit(rustc_driver::catch_with_exit_code(move || {
         rustc_driver::RunCompiler::new(&args, callbacks).run()
-    });
-    std::process::exit(exit_code)
+    }));
 }
```
workingjubilee, line 170: Nice. -80 lines.

workingjubilee (approved):

plrustc/plrustc/src/main.rs
thomcc: Also a lot simpler conceptually. It no longer invokes itself!

PR #232, opened 2023-02-24
Rename `trusted-pgx` to `plrust-trusted-pgx`

eeeebbbbrrrr:
This changeset renames `trusted-pgx` to `plrust-trusted-pgx` so that we can publish it to crates.io while being good citizens.  The intent is for it to be versioned independently of both `pgx` and `plrust`.  It is likely that it'll need to evolve in ways that aren't necessarily tied to specific pgx/plrust releases.

We also introduce a new GUC, `plrust.trusted_pgx_version` so plrust administrators can change the default version.  Which, the default version, is whatever `plrust-trusted-pgx` is set to in this workspace at the time the plrust binary was built.  This value is sussed out by spawning `cargo tree ...` during plrust's new `build.rs` execution.

The `plrust-trusted-pgx` version is now stored in the `pg_proc.prosrc` catalog entry in a new, non-optional, property named "trusted_pgx_version".  This value is not used right now, but it may be necessary in the future should we support re-compilation.

There's now a `./publish.sh` script in the root that only publishes `plrust-trusted-pgx`.


plrust-trusted-pgx/Cargo.toml
```diff
@@ -1,10 +1,9 @@
 [package]
-name = "trusted-pgx"
-# version set to match that of the pgx version it sits on top of, rather than matching plrust's version
-version = "0.7.2"
+name = "plrust-trusted-pgx"
+version = "0.0.0-do-not-use"
 authors = ["TCDI <opensource@tcdi.com>"]
 edition = "2021"
-license = "PostgreSQL Open Source License"
+license = "MIT"
```
eeeebbbbrrrr, line 6: I had to do this to appease publishing to creates.io.

thomcc: you could also do `license-file = "LICENSE.md"` I think? (Possibly might need to copy or symlink the license file into the same directory)

thomcc:
I think using `cargo tree` is not ideal but it's probably the easiest approach.

eeeebbbbrrrr:
> I think using `cargo tree` is not ideal but it's probably the easiest approach.

I don't like it much either, but also didn't like the idea of finding the `plrust-trusted-pgx` directory and directly opening its Cargo.toml and parsing it with the `toml` crate.  That seemed like re-inventing the wheel a bit.

Issue #197, opened 2023-02-04
Improve error message on failure to create function


rustprooflabs:
# Details

I tried creating the following function which fails.  The error message returned is not helpful to figuring out why this did not work.  Can this error message be improved?

```sql
CREATE FUNCTION spi_srf()
    RETURNS TABLE(id BIGINT)
    LANGUAGE plrust
AS
$$
    let query = "SELECT id::BIGINT FROM generate_series(1, 3) id;";

    Spi::connect(|client| {
        let mut results = Vec::new();
        let mut tup_table = client.select(query, None, None)?;

        while let Some(row) = tup_table.next() {
            let id = row["id"].value::<i64>();

            results.push((id));
        }
        Ok(TableIterator::new(results.into_iter()))
    })

$$
;
```

Using `/errverbose` returns:

```
ERROR:  XX000: assertion failed: `(left == right)`
  left: `1`,
 right: `0`
LOCATION:  crating.rs:73
```

workingjubilee:
...Yes indeed, the error message can be improved. However, it is probably not your fault. The assertion that you just hit was never expected to be hit and suggests a bug in PL/Rust occurred.

PR #235, opened 2023-02-24
address issue #197 by blocking RETURNS TABLE, OUT and INOUT function argument

eeeebbbbrrrr:
We don't support these things yet, and this improves the error message from

```
ERROR:  assertion failed: `(left == right)`
  left: `1`,
 right: `0`
```

to

```
ERROR:  not yet implemented: OUT arguments   # (or `RETURNS TABLE functions` or `INOUT arguments`)
```

### DIFF
--- a/plrust/src/pgproc.rs
+++ b/plrust/src/pgproc.rs
@@ -126,6 +126,11 @@ impl PgProc {
             .unwrap_or_default()
     }
 
+    pub(crate) fn proargmodes(&self) -> Vec<i8> {
+        self.get_attr(pg_sys::Anum_pg_proc_proargmodes)
+            .unwrap_or_default()
+    }
+
     pub(crate) fn prorettype(&self) -> pg_sys::Oid {
         // SAFETY:  `prorettype` has a NOT NULL constraint
         self.get_attr(pg_sys::Anum_pg_proc_prorettype).unwrap()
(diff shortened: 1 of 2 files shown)


thomcc (approved):
LGTM.

thomcc:
Not ideal to panic but hopefully we handle it fine and don't poison locks or anything.

eeeebbbbrrrr:
it's converted into a Postgres ERROR that aborts the transaction, using pgx panic handling routines, so it's fine from that perspective.

thomcc:
Right, I just mean if we have a `let x = mutex.lock().unwrap();` somewhere, and `panic!` across it, it will cause problems. That can be fixed by using `let x = mutex.unlock().unwrap_or_else(|e| e.into_inner());` though.

eeeebbbbrrrr:
> Right, I just mean if we have a `let x = mutex.lock().unwrap();` somewhere, and `panic!` across it, it will cause problems. That can be fixed by using `let x = mutex.unlock().unwrap_or_else(|e| e.into_inner());` though.

Oh, I gotcha.  We don't use any locking primitives in plrust.  There's is an `Rc` somewhere, but that'll be fine since we properly unwind the stack.

eeeebbbbrrrr:
yeah, this is gonna fail CI b/c I just merged #232  It's fine  It passes locally

PR #237, opened 2023-02-27
Remove some stale documentation from plrustc's readme, and add mention of some of the env vars it uses to the main readme

thomcc:
Most of the `plrustc` readme was written when things were in an intermediate state that looked pretty different from where we ended up. I've completely ditched the installation guide (it was stale, and we should document that in the main README instead), and (I think) removed all the inaccuracies.

I've also added mention of the environment variables `plrustc` uses to find the sysroot. They're basically the same ones that miri, clippy, etc. might use, although with `PLRUSTC_SYSROOT` acting as the override var (rather than something like `MIRI_SYSROOT`, which is clearly not ours to use).

### DIFF
--- a/plrustc/README.md
+++ b/plrustc/README.md
@@ -30,7 +30,7 @@ Similar to `rustc`, `plrustc` is usually not invoked directly, but instead throu
 
 ## Details
 
-Some additional details are provided for users who intend to run PL/Rust and plrustc under restricted environments via seccomp and/or selinux. These details are subject to change, although if that occurs it will be noted in the changelog.
+Some additional details are provided for users who intend to run PL/Rust and plrustc under restricted environments via seccomp and/or SELinux. These details are subject to change, although if that occurs it will be noted in the changelog.
 
 ### Sysroot configuration
 
--- a/plrustc/README.md
+++ b/plrustc/README.md
@@ -4,8 +4,6 @@
 
 Just to give a brief overview: trusted mode combines various lints with a [custom standard library](https://github.com/tcdi/postgrestd) in an effort to meet the PostgreSQL requirements for a trusted language handler, mainly restricting all access to the underlying operating system. Note that this is only done for the user crate and not dependencies, although dependencies can be disabled in PL/Rust's configuration (and you probably want to do so if you are using "trusted" mode).
 
-FIXME(thom): Okay I rewrote a bunch of stuff and a lot of this is somewhat less accurate. The builder is gone for now, as is, temporarially, the use of RUSTC_WRAPPER. I think we probably *do* need the builder, but.... let's see if we can replace it with something less monsterous????
-
 ## Developing
 
 It's recommended that you use rustup for your toolchain during development. Configure things as specified in the rust-toolchain.toml at the repository root.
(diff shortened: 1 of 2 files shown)


plrustc/README.md
```diff
@@ -26,56 +24,26 @@ We have a few independent workspaces. My `rust-analyzer` configuration currently
...
-5. The compile-time environment is checked.
+3. If `rustc` is on the path, then `rustc --print sysroot` is invoked and that value is used.
+4. The compile-time environment is checked.
    1. First, for `PLRUSTC_SYSROOT` and `SYSROOT` in that order of preference.
    2. Then, for rustup, if both `RUSTUP_HOME` and `RUSTUP_TOOLCHAIN` are set in the environment at runtime, then we will use the path `$RUSTUP_HOME/toolchains/$RUSTUP_TOOLCHAIN` as the sysroot.
-6. If none of these were successful, an error is emitted and compilation will fail.
-
-It's likely that a future version of plrustc will refine this to allow more control. In the short term this is impossible, howrever.
-
+5. If none of these were successful, an error is emitted and compilation will fail.
```
workingjubilee, line 47: We do have the _option_ of exploiting the fact that Markdown demands enumeration incrementing on render and simply numbering all of these with `1.` if we expect the rich text form to be what is read most of the time.

I stress "option" because it does not excite me as an idea.

plrustc/README.md
```diff
@@ -26,56 +24,26 @@ We have a few independent workspaces. My `rust-analyzer` configuration currently
...
-1. If a `-Zplrustc-sysroot=...` argument is present in `PLRUSTC_FLAGS` then this is used. FIXME(thom): this is not true anymore.
-2. If a `--sysroot` argument is provided via normal program arguments, then that value is used.
-3. The runtime environment is checked.
+1. If a `--sysroot` argument is provided via normal program arguments, then that value is used.
+2. The runtime environment is checked.
    1. First, for `PLRUSTC_SYSROOT` and `SYSROOT` in that order of preference.
    2. Then, for rustup: If both `RUSTUP_HOME` and `RUSTUP_TOOLCHAIN` are set, then we will use the path `$RUSTUP_HOME/toolchains/$RUSTUP_TOOLCHAIN` as the sysroot.
-4. If `rustc` is on the path, then `rustc --print sysroot` is invoked and that value is used.
-5. The compile-time environment is checked.
+3. If `rustc` is on the path, then `rustc --print sysroot` is invoked and that value is used.
```
workingjubilee, line 43: I'm guessing that this has the usual interaction in PL/Rust's handler with the GUC configuring the effective PATH, so that this PATH is then what is seen by plrustc?

plrustc/README.md
```diff
@@ -26,56 +24,26 @@ We have a few independent workspaces. My `rust-analyzer` configuration currently
...
-
-### Environment variables
-
-Here are some `plrustc`-specific environment variables that PL/Rust or `plrustc` itself may set. This is in addition to the ones mentioned in "Sysroot Configuration" below.
-
-The `PLRUSTC_FLAGS` variable contains flags passed by PL/Rust which are specific to `plrustc` and required for it to function. The `PLRUSTC_PASSTHROUGH` argument is also used to invoke `plrustc` as if it were plain `rustc`, which it may . FIXME(thom): `PLRUSTC_FLAGS` are not used right now.
-
-PL/Rust is likely to invoke `plrustc` as the value of both `RUSTC` and
-`RUSTC_WRAPPER`. FIXME(thom): seems like we might only need to set it as `RUSTC` now? Not sure.
+Some additional details are provided for users who intend to run PL/Rust and plrustc under restricted environments via seccomp and/or selinux. These details are subject to change, although if that occurs it will be noted in the changelog.
```
workingjubilee, line 33:
```suggestion
Some additional details are provided for users who intend to run PL/Rust and plrustc under restricted environments via seccomp and/or SELinux. These details are subject to change, although if that occurs it will be noted in the changelog.
```

workingjubilee (approved):

plrustc/README.md
thomcc: I have an [editor plugin](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) that handles the numbering for me so it doesn't bother me at all.

plrustc/README.md
thomcc: Yes, it is.

plrustc/README.md
workingjubilee: Neat!

eeeebbbbrrrr (approved):

PR #238, opened 2023-02-28
Fix failing uitests and add them to CI

thomcc:
Somehow this started failing. Add these to CI to prevent that from happening again.

CC @BradyBonnette


thomcc:
This needs to be in the final release, since without this we don't catch the leaky functions (`mem::forget` and such)

eeeebbbbrrrr (approved):

PR #239, opened 2023-02-28
Remove pointless build.rs from plrustc

thomcc:
This was not needed after the overhaul. It's harmless, but makes builds very slightly slower (plrustc already builds so fast that it basically doesn't matter at all).

We should wait for #238 before landing this, since that adds plrustc CI. This does not need to be in the next release.


eeeebbbbrrrr (approved):
Less moving parts is always good!

PR #240, opened 2023-02-28
Add `./build.sh install` as a method of installing `plrustc` (uses `cargo install` to perform the installation)

thomcc:
As described. If no args are given (or the arg isn't literally `install`), the current way is used.

It's likely some additional sophistication for this script wouldn't be amiss, but for now this seems fine.

### DIFF
--- a/plrustc/build.sh
+++ b/plrustc/build.sh
@@ -64,15 +64,25 @@ fi
 
 echo "plrustc build starting..." >&2
 
-env RUSTFLAGS="$RUSTFLAGS" \
-     "$CARGO" build --release \
-        -p plrustc --bin plrustc \
-        --target "$host"
 
-cd "$repo_root"
+if [[ "$1" = "install" ]]; then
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr (approved):
@thomcc should we publish this to crates.io?  `cargo install plrustc` is a neat idea.

thomcc:
It's a neat idea but there's not really a way to make it work. You need to install via the `build.sh`. In particular, this line is load-bearing: https://github.com/tcdi/plrust/blob/45861fe2f29e4822ed778b6d7494aca32d7f6b8a/plrustc/build.sh#L57

If you `cargo install` without setting the rpath, we aren't able to locate the rustc_driver libraries. Sadly, [`rpath = true`](https://doc.rust-lang.org/cargo/reference/profiles.html#rpath) in the cargo profile is also not viable, since it's a boolean and not a path (not to mention, we have to compute that path dynamically...)

PR #241, opened 2023-02-28
Remove `RUSTC_WRAPPER` from the env if set

thomcc:
Meant to do this a while ago. This may or may not clear up the issues we had with sccache (no idea), but we shouldn't inherit this from the user env.


eeeebbbbrrrr (changes requested):
If you will, please document these in the top-level README.md?  It has a list of environment variables that we mess with, and why.

eeeebbbbrrrr (approved):

PR #243, opened 2023-03-03
Enable a few non-plrust lints by default, and add a lint for std::io::{stdout, stderr, stdin}

thomcc:
Went through the builtin lint list looking for things we want to forbid. Justifications in the LINTS.md doc.


thomcc:
I added the lint we want for `std::io::std*` to this too. Not because it's related at all, but because it would conflict and I don't trust myself to not mess up the rebase while on cold medicine.

eeeebbbbrrrr (approved):
Thanks for changing the name of that lint to `plrust_stdio`.

PR #247, opened 2023-03-04
Add GitHub workflow to deploy mdbook docs to GitHub pages

rustprooflabs:
Here's an example of deploying mdbook to GitHub pages.  This may or may not be the path that is decided on for deployment.  Using this requires minor configuration changes to the project itself.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,36 @@
+# Based on https://github.com/rust-lang/mdBook/wiki/Automated-Deployment%3A-GitHub-Actions
+name: Deploy mdbook
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


eeeebbbbrrrr:
What are the things we need configure on the repo?  I probably have the permissions to do all of that...

rustprooflabs:
I think it was just setting up the source/branch options under Settings > Pages and adding the workflow for the action.  The workflow commits the built docs to a `gh-pages` branch, that name can be changed if you want.


image

eeeebbbbrrrr (commented):
LGTM?  Lets try this and see if it works

eeeebbbbrrrr (approved):

PR #248, opened 2023-03-04
take a look at using `#![no_std]` for user functions.

eeeebbbbrrrr:
In terms of getting `#![no_std]` tied in, this pretty much does it.

I dunno if there's more we should look at regarding error handling.  What little bit is here (the `PlRustFunctionError`) type seems to do the needful and none of the existing tests needed to be updated to account for it, which was my primary goal.

### DIFF
--- a/plrust/src/tests.rs
+++ b/plrust/src/tests.rs
@@ -1046,6 +1046,26 @@ mod tests {
         assert_eq!("operation not supported on this platform", &string);
         Ok(())
     }
+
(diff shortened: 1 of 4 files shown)


eeeebbbbrrrr:
If we like this idea I'll update the README before merge.

thomcc:
This needs a working version of #245.

eeeebbbbrrrr:
We decided offline that this just isn't worth the effort.

PR #250, opened 2023-03-06
Use custom FileLoaders to sanitize `include` error messages

thomcc:
This is a few parts, and was pretty involved.

1. First, `plrust` passes `plrustc` information about the crate being compiled via the environment, specifically the crate name (in `PLRUSTC_USER_CRATE_NAME`), and a list of directories that it's expected to access for source files (in `PLRUSTC_USER_CRATE_MAY_ACCESS`).

	This is pretty coarse-grained, and is just a colon-delimited path list. We could consider something more robust if it would be useful.

2. Then, `plrustc` detects when the crate it's compiling happens to be the one given by `PLRUSTC_USER_CRATE_NAME`. If it is, it requires that the paths be subdirectories of the ones given in `PLRUSTC_USER_CRATE_MAY_ACCESS`.

3. Finally, Regardless of the crate being compiled, any `io::Error`s that come from reading files are now always reported as file not found (e.g. `ENOENT`). This is mainly to avoid this sort of issue happening in the future.

To fix this issue for `include_bytes!` we will also need https://github.com/tcdi/postgrestd/pull/39, since `FileLoader` does not (yet) support binary files.

I added a new dependency on `omnipath` to help make paths absolute, even in the case where they don't exist. It's maintained by one of the other maintainers of the rust stdlib, is part of an (unfinished) proposal for new stdlib APIs, and I've given it a read-through and it seems reasonable. I'm not sure it's exactly what we want, but if it's wrong it should be overly-strict rather than overly-permissive.

~~Still needs tests.~~ Doesn't fix the related issue with the `env` macros (might require error matching...).


thomcc:
This needs https://github.com/tcdi/postgrestd/pull/39.

plrustc/plrustc/src/main.rs
```diff
@@ -95,6 +196,63 @@ fn arg_value<'a, T: AsRef<str>>(args: &'a [T], find_arg: &str) -> Option<&'a str
     None
 }
 
+struct ErrorHidingFileLoader;
+
+fn replacement_error() -> std::io::Error {
+    // Unix-ism, but non-unix could use `io::Error::from(ErrorKind::NotFound)`
```
workingjubilee, line 202: Let's just say Postgres itself has a lot of portability shims aaand well, it tends to shim other systems into looking more like Unix rather than the other way around.

thomcc:
Crap, I think I just realized an issue here.

plrust/Cargo.toml
```diff
@@ -60,6 +60,7 @@ prettyplease = "0.1"
 syn = "1"
 quote = "1"
 proc-macro2 = "1"
+omnipath = "0.1.1"
```
workingjubilee, line 63: I glanced at this and the implementation seems POSIXly correct, yeah, and the library is sound (and also more annoying to maintain ourselves because it's one of those extremely pedantic things by nature).

README.md
```diff
@@ -293,6 +293,8 @@ overridden through configuration.
...
 | PGX_PG_CONFIG_INCLUDEDIR-SERVER             | Provided by the running Postgres instance                                     | Used by `pgx` to build the PL/Rust user function.                                                                                                                                                                     |
+| PLRUSTC_USER_CRATE_NAME                     | The name of the user crate                                                    | Used by `plrust` to help `plrustc` distinguish between the PL/Rust user function and dependencies.                                                                                                                    |
+| PLRUSTC_USER_CRATE_MAY_ACCESS               | The user crate dir and the cargo target dir, separated by `:`.                | When compiling the user crate (given by `PLRUSTC_USER_CRATE_NAME`), `plrustc` ensures no source files outside of these (`:`-delimited) directories are accessed. Every path in the list should be absolute and have symlinks resolved. The sysroot is currently automatically added to this list. |
```
workingjubilee, line 297: For a hot second I thought this was going to be a bool, while reading the rest of the code. Could this name suggest the PATH-like nature of it?

plrust/Cargo.toml
thomcc: It's actually wrong to use in plrustc, I think, since we should only ever allow canonicalized paths. Annoying.

thomcc: Now only plrust (not plrustc) uses it, where it seems fine.

eeeebbbbrrrr (approved):
feel free to merge after CI passes.

The code looks fine.

I'm not thrilled we have to do this and wish there was a different way, but if there is, it's not obvious at the moment.

Commits like https://github.com/tcdi/plrust/pull/250/commits/d277e0b1e3a425c07c168df5d5a3cfef1f83daad make me nervous only because this kind of thing is incredibly hard to get right.  Definitely glad you realized this here in the PR!

thomcc:
> Commits like https://github.com/tcdi/plrust/commit/d277e0b1e3a425c07c168df5d5a3cfef1f83daad make me nervous only because this kind of thing is incredibly hard to get right. Definitely glad you realized this here in the PR!

Yeah, I agree :(.

PR #251, opened 2023-03-07
Avoid linking to `rust-lang/rust`'s issue tracker if `plrustc` panics

thomcc:
I wrote some buggy code and noticed we tell the user to file a bug against `rust-lang/rust`. That seems bad since it will probably annoy people upstream if they have to field plrustc issues.

Most of the complexity here is:

1. Allowing this to be mor tested.
2. Setting up the code needed to report a one-shot error from a location that doesn't have access to the error reporting machinery.

The second of these is the more confusing bit, but it's also something we will want anyway in the long term (at which point I'll factor it out).

### DIFF
--- a/plrustc/plrustc/src/lints.rs
+++ b/plrustc/plrustc/src/lints.rs
@@ -357,8 +357,38 @@ fn match_def_path<'tcx>(cx: &LateContext<'tcx>, did: DefId, syms: &[&str]) -> bo
         .eq(path.iter().copied())
 }
 
+// Used to force an ICE in our uitests. Only enabled if
+// `PLRUSTC_INCLUDE_TEST_ONLY_LINTS` is enabled in the environment, which we do
+// explicitly in the tests that need it.
+declare_lint! {
+    pub(crate) PLRUST_TEST_ONLY_FORCE_ICE,
+    Allow,
+    "This message should not appear in the output"
+}
+
+declare_lint_pass!(PlrustcForceIce => [PLRUST_TEST_ONLY_FORCE_ICE]);
+
+impl EarlyLintPass for PlrustcForceIce {
+    fn check_fn(
+        &mut self,
+        _: &EarlyContext<'_>,
+        fn_kind: ast::visit::FnKind<'_>,
+        _: Span,
+        _: ast::NodeId,
+    ) {
+        use ast::visit::FnKind;
+        const GIMME_ICE: &str = "plrustc_would_like_some_ice";
+        if matches!(&fn_kind, FnKind::Fn(_, id, ..) if id.name.as_str() == GIMME_ICE) {
+            panic!("Here is your ICE");
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


workingjubilee:
I believe you can use a source file with the expression `break rust;` to kill the compiler rather than a bunch of linting and environment machinery.

workingjubilee:
If we need a different exact internal handling pattern for this, then rather than a bunch of configuration, we should just enable `break plrust;`

thomcc:
> I believe you can use a source file with the expression break rust; to kill the compiler rather than a bunch of linting and environment machinery.

I tried that, it doesn't actually cause a panic.

thomcc:
> If we need a different exact internal handling pattern for this, then rather than a bunch of configuration, we should just enable `break plrust;`

This is a lot more code, and feels less robust (specifically):

```rs
impl EarlyLintPass for PlrustcForceIce {
    fn check_expr(&mut self, _: &EarlyContext<'_>, expr: &ast::Expr) {
        let ast::ExprKind::Break(None, Some(breakee)) = &expr.kind else {
            return;
        };
        let ast::ExprKind::Path(None, path) = &breakee.kind else {
            return;
        };
        if path.segments.len() != 1 {
            return;
        }
        if path.segments[0].ident.name.as_str() == "plrust" {
            panic!("It looks like you're trying to `break plrust;` would you like some ICE?");
        }
    }
}
```

Note that `break rust` isn't a lint, its a genuine compile error that is emitted during HIR lowering. While it's *possible* for us to add an analysis pass to do stuff like that, it would be extremely overkill.

workingjubilee:
Ah, alas.

thomcc:
Also note that if we do this, it still ends up with an error like
```
error[E0425]: cannot find value `plrust` in this scope
  --> $DIR/break_rust.rs:8:11
   |
LL |     break plrust;
   |           ^^^^^^ not found in this scope

thread 'rustc' panicked at 'It looks like you're trying to `break plrust;` would you like some ICE?', plrustc/src/lints.rs
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: internal compiler error: unexpected panic

note: `plrustc` unexpectedly panicked. This is probably a bug.

note: Please file a bug report at <https://github.com/tcdi/plrust/issues/new>

note: plrustc version: <version here>

error: aborting due to previous error

For more information about this error, try `rustc --explain E0425`.
```

It does still run the early lint pass (so our error happens), but I'm not really a fan of that, since I'm intending on eventually enforcing that our (current) uitests are compile-fail, and would like it so that they don't have unrelated errors in the output.

workingjubilee:
Nah, the current is fine. What I was more thinking about is that there's no actual concern re: conditionalization, since there's no real threat from a user panicking their own compilation. We're more likely to accidentally use a conditional to shut off a real lint because we were editing code at... 1 in the morning, apparently.

eeeebbbbrrrr (approved):
I'm assuming you and @workingjubilee came to a consensus here...

workingjubilee:
Yes.

PR #253, opened 2023-03-08
Fix several issues with lint application and detection

### LINKED ISSUE #233: `plrust_print_macros` lint fires multiple times for each violation
This lint landed with an issue where each violation is reported multiple times (we landed it with this issue knowingly, since it's still better than not reporting it, and would like it in the release).

You can see this here: https://github.com/tcdi/plrust/blob/c2fe02b764e80eea00aa4aee42bc312551729059/plrustc/plrustc/uitests/print_macros.stderr

The issue is that every time we walk over the expression, we walk the full expansion tree (e.g. the full macro backtrace), which touches this multiple times.

I think we can fix it by stopping one past the first macro defined outside the user crate, but it's tricky to know what that is, so perhaps stopping as soon as we hit `std` would be easier.

thomcc:
This fixes several significant issues with lint application and detection.

1. Expansions from external crates are now linted, because otherwise things like `pgx::log!("{:?}", <bad thing here>)` (or `let _ = std::format_args!("{:?}", bad_thing())`) could fail to be detected, especially if `<bad thing here>` involves a macro.

2. More significantly, macro detection is much more robust. This fixes several holes in the detection and also fixes #233. (That said, some of the macro detection code could potentially be replaced by macro overrides soon).

This is applied on top of #251 in order to avoid additional merge issues (our tests currently suck and I'm tired).


eeeebbbbrrrr (approved):
I can't review this for correctness b/c I know nothing about how a rustc_driver is supposed to be put together.

But I don't see any glaring problems here.  :)

thomcc:
> I can't review this for correctness b/c I know nothing about how a rustc_driver is supposed to be put together.

Yeah, so honestly sanity checking the uitests vs their stderr output is probably worth doing as a rubberstamp review. TBH, I had not been careful enough about it — at least one (maybe more?) of these issues that this PR fixes would have been caught if I had looked more carefully at the `.stderr` files.

For example: if you compare this test (links are to current `main` without this PR): https://github.com/tcdi/plrust/blob/1d41038905dca5e34b49181277f4a0f99ec1d6ec/plrustc/plrustc/uitests/fs_macros.rs with this output: https://github.com/tcdi/plrust/blob/1d41038905dca5e34b49181277f4a0f99ec1d6ec/plrustc/plrustc/uitests/fs_macros.stderr. The test has cases `_[A-J]` but stderror only complains about `_A`, `_B`, `_C`, `_D` and `_H`, meaning the lint misses `_E`, `_F` and `_G`... 🤦. I even thought to add the test for those situations too, so this is pretty clownshoes.

(That said, note that there are sadly still some things we can't check in uitests, since the uitests don't have `postgrestd`, but I wouldn't put that stuff in a test file)

thomcc:
No rush on that, I'm done for the day, and I'm not worried about it on this PR either. Just saying, even without understanding the rustc_driver stuff, some review for obvious mistakes like that is still appreciated.

Issue #255, opened 2023-03-09
Improve error - Create function with anonymous variables


rustprooflabs:
# Details

PL/Rust functions are not expected to support anonymous input parameters like other languages do, e.g `$1`. The current error message when this is attempted is not helpful to the end user.

```sql
CREATE FUNCTION plrust.strlen(TEXT)
    RETURNS BIGINT
    LANGUAGE plrust STRICT
AS $$
    Ok(Some(arg0.len() as i64))
$$;
```

Returns

```
ERROR:  assertion failed: `(left == right)`
  left: `0`,
 right: `1`
DETAIL:  
```

It would be great to have a more informative error message in this case.

eeeebbbbrrrr:
@rustprooflabs, got a PR up for this at https://github.com/tcdi/plrust/pull/258

The rules are:

- error if the argument is unnamed
- error if the argument name isn't also a valid Rust identifier

I'll let you update the docs for this.  The docs probably need a section about the rules and regulations such as, now, argument naming, argument types (`Option<T>` or `T` depending on STRICT), return types (`Result<Option<T>>`), and maybe other things I haven't thought of.

PR #258, opened 2023-03-09
Fix issue #255

eeeebbbbrrrr:
PL/Rust does not support unnamed function arguments.  And more generally, PL/Rust does not support function argument names that aren't also valid Rust identifiers.

This catches all of these situations and raises the appropriate SQL error.

### DIFF
--- a/plrust/src/user_crate/crate_variant.rs
+++ b/plrust/src/user_crate/crate_variant.rs
@@ -9,7 +9,6 @@ Use of this source code is governed by the PostgreSQL license that can be found 
 use crate::{user_crate::oid_to_syn_type, PlRustError};
 use eyre::WrapErr;
 use pgx::PgOid;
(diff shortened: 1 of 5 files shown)


plrust/src/user_crate/crate_variant.rs
```diff
@@ -54,10 +52,6 @@ impl CrateVariant {
                 }
             };
 
-            let arg_name = match maybe_arg_name.as_deref() {
-                Some("") | None => Ident::new(&format!("arg{}", idx), Span::call_site()),
-                Some(argument_name) => Ident::new(&argument_name.clone(), Span::call_site()),
-            };
```
workingjubilee, line 60: Ahh, this was what I missed while looking for the name expansion.

workingjubilee (approved):
Nice!

PR #261, opened 2023-03-13
Adds workflow_dispatch to CI

BradyBonnette:
This should provide us with the ability to kick off a build whenever, similar to PGX. The ability will only show up once this gets into `main` however


BradyBonnette:
Not sure what's going on with GHA, but that "running" aarch64 instance is actually done.

eeeebbbbrrrr (approved):

PR #262, opened 2023-03-14
prepare v1.0.0-rc.1

eeeebbbbrrrr:


eeeebbbbrrrr:
I expect CI to fail on this b/c `plrust-trusted-pgx:1.0.0-rc.1` isn't yet on crates.io, but this will want it there.

PR #265, opened 2023-03-21
Add a lint for `'static` impl shenanigans

thomcc:
~~This applies on top of #264 because I'm lazy.~~ Edit: not anymore

It basically works but needs more testing and docs. I've decided to handle it by detecting the `impl ... for <something that contains 'static in here>` pattern, which should be pretty rare.


thomcc:
Okay, I'm pretty confident this is the right approach (and that trying to solve it by additional analysis passes is more trouble than its worth).

thomcc:
I think the CI failure is just aws flakiness. (CC @BradyBonnette)

workingjubilee:
Yes. Will rerun those tests when it settles.

plrustc/plrustc/uitests/static_impl_unsound.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use std::fmt::Display;
+
+trait Displayable {
+    fn display(self) -> Box<dyn Display>;
+}
+
+// This is more complex than the one in the issue, to make sure the `Box`'s
+// lang_item status doesn't bite us.
+impl<T: Display> Displayable for (T, Box<Option<&'static T>>) {
```
workingjubilee, line 11: This is making me wonder: `Box<T>` is a lang item, but more than that, it is `#[fundamental]`. Should we have any special concerns about other `#[fundamental]` types here?

plrustc/plrustc/src/lints.rs
```diff
@@ -62,6 +62,94 @@ impl<'tcx> LateLintPass<'tcx> for NoExternBlockPass {
...
+                polytrait.iter().any(|poly| {
+                    self.segments_have_static(poly.trait_ref.path.segments)
+                })
+            }
+            // Something like `Vec<T>` or `Option<T>`. Need to look inside...
+            TyKind::Path(qpath) => {
+                let segs = match qpath {
+                    hir::QPath::Resolved(Some(maybe_ty), _) if self.has_static(maybe_ty) => return true,
+                    hir::QPath::TypeRelative(t, _) if self.has_static(*t) => return true,
+                    hir::QPath::LangItem(..) => return false,
```
workingjubilee, line 130: Or is this why you mentioned the `Box<Option<&'static T>>` thing?

plrustc/plrustc/src/lints.rs
```diff
@@ -62,6 +62,94 @@ impl<'tcx> LateLintPass<'tcx> for NoExternBlockPass {
...
+
+    fn segments_have_static(&self, segs: &[hir::PathSegment]) -> bool {
+        segs.iter().any(|seg| {
+            seg.args().args.iter().any(|arg| match arg {
+                hir::GenericArg::Lifetime(hir::Lifetime {
+                    res: hir::LifetimeName::Static,
+                    ..
+                }) => true,
+                hir::GenericArg::Type(t) => self.has_static(t),
+                _ => false,
```
workingjubilee, line 147: I know we're just dumping all of these on the floor because we don't care but could we use exhaustive match and discard the inner match, here? It makes it clearer what we're ignoring.

```suggestion
                hir::GenericArg::Const(_) | hir::GenericArg::Infer(_)
                // Wasn't static
                | hir::GenericArg::Lifetime(_) => false,
```

workingjubilee: Also it's not approximately 500 variants, unlike the other enums in rustc.

plrustc/plrustc/uitests/static_impl_unsound.rs
thomcc: Not as far as I know.

plrustc/plrustc/src/lints.rs
thomcc: Yeah, but I think it only happens if the path refers to a function.

plrustc/plrustc/uitests/static_impl_unsound.rs
thomcc: To be more concrete, we are looking somewhat blindly for `'static` through almost everything, so fundamental or not, it doesn't make a huge difference.

workingjubilee (approved):

PR #269, opened 2023-03-22
Add an `extern crate` lint

thomcc:
Fixed version of #245. Not a soundness fix, not sure if we should do it, but it's an option.

### DIFF
--- a/doc/src/config-lints.md
+++ b/doc/src/config-lints.md
@@ -192,6 +192,15 @@ mod bar;
 mod baz;
 ```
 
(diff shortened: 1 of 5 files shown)


thomcc:
I'm guessing this fails because we have `extern crate` somewhere in a macro. Which is probably an indication we shouldn't do this.

thomcc:
I'll close this but keep the branch alive for if we ever want to address postgrestd by just doing `#![no_std]` instead.

PR #271, opened 2023-03-23
Move lints into their own file

thomcc:
This changes no logic just where code lives.


workingjubilee (approved):
But don't you _like_ having thousands of lines of code in a single file which you have to scroll forever to find any issues in?

PR #273, opened 2023-03-23
Only set `-Clink-args=-Wl,-undefined,dynamic_lookup` on macos

thomcc:
This has bugged me for a billion years. This flag does nothing on non-macOS platforms, and it's honestly surprising that it isn't an error on other linkers.


workingjubilee (approved):
I kept forgetting to do this because I kept thinking we had already done this because we had done it but in other repos.

PR #274, opened 2023-03-26
Docs: Intro, Cross compile install, and Configuration cleanup 

rustprooflabs:
Another docs related PR to fill out the earlier sections.  The first commit moves more content from main project's README.md into the new documentation.  This fills out the documentation's home page with more of the big concepts. 

The second commit focused on adding cross compile installation steps and making the steps/differences more clear.  The installation process was split into a dedicated "prereqs" section before the PL/Rust installation section. This makes the 3 installation paths (untrusted, trusted, trusted + cross compile) more obvious in the Install PL/Rust section.

The cross compile configuration notes are left behind in the 3rd section.  That content will be moved into the main configuration section, probably in a new PR targeting that section.

image

The easiest way to preview the changes to the `mdbook` content is checking out the branch and serving the book locally.  See [docs/README.md](https://github.com/tcdi/plrust/blob/4b1f1e172dd7925da12d3392bb50544c2ff1d744/doc/README.md)


rustprooflabs:
I like the way the PL/Rust Installation sections reads now.

image

rustprooflabs:
3rd commit cleans up the configuration page.  The table with did not render well with how wide it was.  Expanded details into section based format.

doc/src/config-pg.md
```diff
@@ -1,49 +1,160 @@
...
+```bash
 plrust.work_dir = '/tmp'
 ```
 
 
-Failure to set these required variables will cause `plrust` extension to not function.
+## Lints
 
-The PL/Rust-specific configuration options are in the following table.
+#### `plrust.compile_lints` (string)
```
thomcc, line 44: We should probably discourage users from tweaking this (or at least note that doing so can completely compromise the "trusted"-ness of plrustc).

I think if they tweak it it also prevents us from updating it in new versions?

rustprooflabs: :+1: Same with `plrust.required_lints`?

workingjubilee: Yes, required lints are, well, required.

rustprooflabs: Yeah, that makes sense. Thanks!

I moved the lints config section to the end of the page and added a warning to not adjust them.

rustprooflabs:
After adding the warning about not changing the lints, the Under the Hood section felt more appropriate for the larger Lints section. The env vars section is another thing most users shouldn't need to change.

Before

image

After

image

PR #281, opened 2023-04-06
fail to create on databases that aren't `UTF8`

eeeebbbbrrrr:
PL/Rust can't support databases that aren't UTF8 because of Rust's requirements around Strings being proper UTF8.  This PR causes `CREATE EXTENSION plrust` to raise an ERROR in this case:

```sql
[v14.6][2495941] plrust=# create extension plrust;
ERROR:  PL/Rust only supports UTF8-encoded databases
```

This might be a little heavy-handed now but it gives us time to consider other options for relaxing this requirement.

The decision here stems from the fact that, currently, `pgx` blindly assumes all `TEXT` Datum conversions are valid UTF8, when they might not be.  That's a problem too, but it's still not clear the RightWay to resolve that and likely wouldn't do us any good here in plrust anyways.

### DIFF
--- a/doc/src/install-plrust.md
+++ b/doc/src/install-plrust.md
@@ -200,9 +200,11 @@ sudo chown root -R /usr/lib/postgresql/15/lib/
 Create a `plrust` database and connect to the `plrust` database
 using `psql`.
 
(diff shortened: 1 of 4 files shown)


workingjubilee (approved):
Reviewed the out-of-band discussion and I think this is good.

PR #286, opened 2023-04-11
Update create function examples to include REPLACE for easy copy/paste.

SeanOzzy:
This is a very simple update to modify the example create function SQL statements to use CREATE OR REPLACE in order to make copy/paste more simple since many examples share the same function schema and name combination.

### DIFF
--- a/doc/src/install-plrust.md
+++ b/doc/src/install-plrust.md
@@ -228,7 +228,7 @@ that simply returns the integer 1.
 
 
 ```sql
-CREATE FUNCTION plrust.one()
+CREATE OR REPLACE FUNCTION plrust.one()
     RETURNS INT LANGUAGE plrust
 AS
 $$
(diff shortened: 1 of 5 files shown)


eeeebbbbrrrr (approved):
Thanks!

We'll be working on docs for quite a bit longer, so anything like this is greatly appreciated.

eeeebbbbrrrr:
I know we ask for PRs to go against `develop`, but this one is fine to go straight to `main`.  It's from `main` that the `tcdi.github.io/plrust` site is updated, and I see no harm in getting this out there now.

Thanks for the contribution, @SeanOzzy.  This is the first contribution since we released v1.0.0.  Congrats!

workingjubilee:
I'm not trying to block this from landing but if the main motivation is "they have the same name, thus collisions happen", should we change the names?

eeeebbbbrrrr:
I'm going to merge this into `develop` and cherry-pick into `main`.

PR #287, opened 2023-04-11
Adds Dockerfile for quick trial use of plrust

BradyBonnette:
Because plrust is a bit convoluted to build and run, we are introducing a Dockerfile that, when ran, should containerize everything necessary with one command so that anyone can test plrust locally. This should not ever be used for any production system, as the Postgres server being set up in this container has essentially 0 security considerations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,7 +147,7 @@ jobs:
       run: cd plrustc && ./build.sh && cp ../build/bin/plrustc ~/.cargo/bin
 
     - name: Run cargo pgx init for PG14
-      run: cargo pgx init --pg14 $(which pg_config)
+      run: cargo pgx init --pg$PG_VER $(which pg_config)
 
     - name: Test PL/rust as "untrusted"
       if: matrix.target == 'host'
@@ -155,7 +155,7 @@ jobs:
 
     - name: Test PL/rust as "trusted" (inc. postgrestd)
       if: matrix.target == 'postgrestd'
-      run: cd plrust && STD_TARGETS="aarch64-postgres-linux-gnu" ./build
+      run: cd plrust && STD_TARGETS="aarch64-postgres-linux-gnu" ./build && cargo test --verbose --no-default-features --features "pg$PG_VER trusted"
 
     - name: Print sccache stats (after build)
       run: sccache --show-stats
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,7 +146,7 @@ jobs:
     - name: install plrustc
       run: cd plrustc && ./build.sh && cp ../build/bin/plrustc ~/.cargo/bin
 
-    - name: Run cargo pgx init for PG14
+    - name: Run cargo pgx init
       run: cargo pgx init --pg$PG_VER $(which pg_config)
 
     - name: Test PL/rust as "untrusted"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -323,7 +323,7 @@ jobs:
 
     - name: Test PL/rust as "trusted" (inc. postgrestd)
       if: matrix.target == 'postgrestd'
-      run: cd plrust && STD_TARGETS="x86_64-postgres-linux-gnu" ./build
+      run: cd plrust && STD_TARGETS="x86_64-postgres-linux-gnu" ./build && cargo test --verbose --no-default-features --features "pg$PG_VER trusted"
 
     # # Maybe this is not necessary if we switch to sccache?
     # # Attempt to make the cache payload slightly smaller.
--- a/Dockerfile.try
+++ b/Dockerfile.try
@@ -53,7 +53,7 @@ RUN cd plrustc && ./build.sh && cp ../build/bin/plrustc ~/.cargo/bin
 RUN cargo pgx init --pg15 $(which pg_config)
 
 # Build and install plrust as trusted
-RUN cd plrust && STD_TARGETS="$(uname -m)-postgres-linux-gnu" ./build && cargo pgx install --features trusted
+RUN cd plrust && STD_TARGETS="$(uname -m)-postgres-linux-gnu" ./build && cargo pgx install --release --features trusted
 
 # Reset the permissions of files/directories that was created or touched by the postgres user
 # Switching to the root user here temporarily is easier than installing and setting up sudo
--- a/Dockerfile.try
+++ b/Dockerfile.try
@@ -71,10 +71,8 @@ USER postgres
 RUN pg_createcluster 15 main
 
 # Set up host based auth. *DO NOT* use this in production!
-RUN cat <<EOF > /etc/postgresql/15/main/pg_hba.conf
-local all all trust
-host all all 0.0.0.0/0 trust
-EOF
+RUN echo 'local all all trust' > /etc/postgresql/15/main/pg_hba.conf && \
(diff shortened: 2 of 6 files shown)


Dockerfile.try
```diff
@@ -0,0 +1,110 @@
...
+        jq \
+        libssl-dev \
+        llvm-11 \
+        make \
+        postgresql-server-dev-15 \
+        pkg-config && \
+    rm -rf /var/lib/apt/lists/*
+
+# Set up permissions so that the postgres user can install the plrust plugin
+RUN chmod a+rwx `$(which pg_config) --pkglibdir` `$(which pg_config) --sharedir`/extension
```
eeeebbbbrrrr, line 30: can we restore the permissions after `plrust` is installed?

I know this is just for people to try out, but we might should be good citizens where we can.

eeeebbbbrrrr (changes requested):
Other than the one comment, I'd like to see the book updated with a section on using this.

I think we're just gonna need to make that a default thing we do with PRs going forward.

BradyBonnette:
> Other than the one comment, I'd like to see the book updated with a section on using this.

That was always part of my plan. I just wanted to get something pushed up first so I could test it out in other environments. That is, I wanted everything squared away before I solidified it in documentation.

Dockerfile.try
BradyBonnette: I realized that I misinterpreted what you had asked. I believe I have a solution for this coming up in a commit.

.github/workflows/ci.yml
```diff
@@ -147,15 +147,15 @@ jobs:
       run: cd plrustc && ./build.sh && cp ../build/bin/plrustc ~/.cargo/bin
 
     - name: Run cargo pgx init for PG14
```
eeeebbbbrrrr, line 149: update `- name:`, I suppose now that it's parameterized.

Dockerfile.try
```diff
@@ -0,0 +1,133 @@
...
+        clang \
+        clang-11 \
+        gcc \
+        git \
+        jq \
+        libssl-dev \
+        llvm-11 \
+        make \
+        postgresql-server-dev-15 \
+        pkg-config && \
```
eeeebbbbrrrr, line 26: is there any of this list we can remove?  Like are the `clang*` and `llvm*` packages required?  What about `make` and `jq` and even `gcc`?

eeeebbbbrrrr: hmm... it may be that pgx itself needs most of those.  Probably don't need `jq` tho.  Unless it's used here in the Dockerfile?

Dockerfile.try
```diff
@@ -0,0 +1,133 @@
...
+RUN cat <<EOT >> /etc/postgresql/15/main/postgresql.conf
+shared_preload_libraries='plrust'
+plrust.work_dir='/tmp'
+EOT
+
+# Set up custom prompt
+RUN cat <<EOT >> ~/.psqlrc
+\set PROMPT1 '%/(plrust)=# '
+\set PROMPT2 '%/(plrust)-# '
+EOT
```
eeeebbbbrrrr, line 95: Probably a dumb question, but what does the "T" stand for?  I always write heredocs with EOF where "F" means "file".  Never seen "EOT"

.github/workflows/ci.yml
BradyBonnette: Yeah good point.

Dockerfile.try
BradyBonnette: Yeah I believe most of those are needed for the compilation of various things. `jq` is needed within the Dockerfile itself to parse out the exact version of pgx to use (as opposed to hard-coding a version value and grepping it later): https://github.com/tcdi/plrust/pull/287/files#diff-81320fd3745575634e7102a717f7a28356d7a9c647d5bdd4d43dfbbc4e41c61dR46

Dockerfile.try
BradyBonnette: In reality it doesn't matter, the heredoc delimiter could be specified as `FOOBAR` and it wouldn't make a difference. I think I copied the structure from somewhere else.

If it makes it more readable I can change it to EOF.

Dockerfile.try
BradyBonnette: @thomcc any comment regarding anything that could be removed here?

BradyBonnette: Now that I am re-examining it, `TARGET_VERSION` is a really dumb name for that environment variable. Ill rename it to something more meaningful.

doc/src/try-plrust-with-docker.md
```diff
@@ -0,0 +1,74 @@
+# Try PL/Rust with Docker
+
+If you would like to give PL/Rust a try without succumbing to the long arduous process of installing everything that is required, then this next section is for you!
```
eeeebbbbrrrr, line 3: Lets maybe tone this down a bit.  Yeah, plrust is complex to install, but we don't need to talk it down.  :)

doc/src/try-plrust-with-docker.md
```diff
@@ -0,0 +1,74 @@
...
+docker run -it -p 5432:5432 tcdi/try-plrust server
+```
+
+This will set up everything that is necessary and run the Postgres server only, binding to TCP port 5432. The final `server` argument in the command indicates that it should launch the `server` script upon container bootup.
+
+To exit out of server mode, press Ctrl+c.
+
+## Caveats
+
+* This Docker setup does not take many security precautions into consideration. As such, the way `Dockerfile.try` is constructed should not be considered a best practice as it relates to setting up and securing a Postgres instance with PL/Rust installed. The resulting container that gets built from this should not be considered secure, and thus should not be used in any production environment whatsoever.
```
eeeebbbbrrrr, line 72: I think you buried the lede here.  I'd start with something like:  "This Dockerfile should not be used in a production environment.  It does not take any security precautions"

doc/src/try-plrust-with-docker.md
```diff
@@ -0,0 +1,74 @@
...
+
+* Mapping internal Postgres directories to the host environment through [bind mounts](https://docs.docker.com/storage/bind-mounts/) or [volumes](https://docs.docker.com/storage/volumes/) is a bit tricky since Postgres doesn't like many things owned and ran by root. Because of this, any functions and data created in a session started by `docker run ...` will be destroyed upon container termination.
```
eeeebbbbrrrr, line 74: Similar here:  "The Postgres data directories are not persistent and are destroyed upon container termination.  Externally mounting Postgres' data directories is outside the scope of this example"

doc/src/try-plrust-with-docker.md
```diff
@@ -0,0 +1,74 @@
+# Try PL/Rust with Docker
+
+If you would like to give PL/Rust a try without succumbing to the long arduous process of installing everything that is required, then this next section is for you!
+
+The PL/Rust repository contains a Dockerfile named `Dockerfile.try` that contains everything you need to spin up and test PL/Rust in your own environment.
+
+The following instructions assume that you have a very basic understanding of what [Docker](https://www.docker.com) is and that you already have it installed. If you do not have Docker installed yet, instructions for your particular environment can be found here: <https://docs.docker.com/engine/install/>
```
eeeebbbbrrrr, line 7: In general I think it's best if our docs don't address the reader directly.  So instead of saying, "... assume that you have a...", maybe something like "assume a very basic understanding of Docker".

There's a few instances of this below.  I suppose it's a style thing.  In the docs I just pushed I tried really hard to eliminate talking to the reader.

eeeebbbbrrrr (changes requested):

PR #294, opened 2023-04-18
Updates to use `pgrx`.

eeeebbbbrrrr:
"pgx" was just renamed to `pgrx` and this addresses the mechanical changes here in PL/Rust.

One concern I have, which can be discussed here and addressed in a separate PR is that PL/Rust used to add `use pgx::prelude::*` to the top of every user function.

Now it's `use pgrx::prelude::*`.  This isn't a big deal in general, but more broadly, if existing PL/Rust functions fully-qualify something from plrust-trusted-pgx, say they wrote `pgx::spi::Spi`, then after this PR is released that code would no longer compile and would need to be modified to be `pgrx::spi::Spi`.

I suppose we could add a `use pgrx as pgx` for backwards compatibility, but we'd have to carry this with us forever.


workingjubilee:
Sorry it took so long, I tried to be a bit more attentive to the failure modes we found progressively yesterday and my brain was a slog today and I needed to remember what I needed to read to understand the concern fully and how things work and and and...

So,
> I suppose we could add a `use pgrx as pgx` for backwards compatibility, but we'd have to carry this with us forever.

This won't necessarily work because someone can always fully-qual to the "root", e.g. they could use the form `::pgx::spi::Spi`. In Rust Edition 2015, that form will resolve from a notional "crate root". But when compiling in Rust Edition 2018 or later, that form resolves to an **external** space, the "extern prelude" that includes all the crates, so it always touches the right crate. That's part of why we use that fully qualified form: so users can't interfere with it via similar hijinx. As we can't actually preserve backwards compatibility across all cases, I advise we don't try, as some forms will break on the extern prelude changing.

eeeebbbbrrrr:
> As we can't actually preserve backwards compatibility across all cases, I advise we don't try

That's legitimate for sure.  What, however, might this mean for plrust's next semver?  This is all new territory for us here since plrust is brand new.

I definitely don't want to just jump to v2.0.0.

workingjubilee:
While it definitely might have been slightly smarter to do things the other way around, of settling the PGRX name first, and then PL/Rust 1.0, since there are no semver questions in that case, it happened that we did it this way around.

I think we shouldn't establish the precedent that breaking changes in ~~pgx~~ pgrx necessitate a version bump in PL/Rust.

IMO the only way to actually handle the compat layer in a principled way is, because technically the cdylib functions are independent from PL/Rust, we allow people to configure everything so that they could hypothetically continue to use pgx 0.7.4 even though PL/Rust is built with pgrx but... I don't think we want that, for obvious reasons.

So we can also just take the ~~unprincipled~~ arbitrary route and shrug it off. "Hey, it's a patch fix." Semver, at the end of the day, is a means of communication, it can't be entirely removed from subjectivity, because it doesn't capture the subtlety of build components changing at all.

workingjubilee (approved):
I have both mechanically checked and inspected by eye for any of the problems that arose from the namechange in the main pgrx repo and I believe none of those have affected the code changes here, aside from the more philosophical question regarding the migration strategy.

PR #298, opened 2023-04-24
Function Capabilities & Zero Copy Arrays

eeeebbbbrrrr:
Adds a `capabilities` property to our `pg_proc.prosrc` json structure and defines an enum of known function capabilities, which right now only contains `ZeroCopyArrays`.

If a function is re-compiled (really only via a pg_restore) then we'll use the set of old capabilities and adjust the wrapper code accordingly.

New functions get the `ZeroCopyArrays` capability by default, which uses `pgrx::Array<'a, T>` instead of `Vec<T>`, which old functions would have been using.

Also adds `pgrx::Array` and friends to `plrust-trusted-pgrx`.


BradyBonnette:
Reviewed after merge 👍

PR #309, opened 2023-05-23
Bump compiletest-rs dep

thomcc:
Fixes the issue @BradyBonnette was hitting in CI.


BradyBonnette:
I am going to try this single change in my fork, since I am building out some CI stuff anyways (I can only test the CI changes in my fork). 

If this works, we could either use these changes here in your PR, or we could keep the above changes in my fork and it'd get merged in when I am done. What do you think?

thomcc:
We discussed offline and @BradyBonnette is just going to carry this change in his fork until it merges.

PR #312, opened 2023-05-24
Prepare v1.1.3

eeeebbbbrrrr:
Welcome to PL/Rust v1.1.3.  This release fixes a performance regression using functions with `ARRAY` arguments.

It also no longer compiles UDFs with `debug-assertions`.  This was, unfortunately, a development-only change that found its way into previously released versions.  

While not required, it may be desirable to `CREATE OR REPLACE` any existing `LANGUAGE plrust` functions in order to take advantage of these performance improvements.


thomcc (approved):
LGTM

BradyBonnette (approved):

PR #318, opened 2023-06-01
Add a lint for the `Self` tuple struct soundness hole

thomcc:
See comments and such.


eeeebbbbrrrr:
This presents an interesting conundrum for existing plrust functions that were compiled before this lint. 

A new plrust with this lint will require it of any function it tries to load. 

This is our first new lint since initial release and I guess we didn’t quite sort this situation out ahead of time. 

I’m sure there’s a path forward here to allow existing functions to load, but it’s late here on the east coast. We can figure it out over the next few days.

thomcc:
FWIW this pattern is quite rare, to the point where a crater run didn't find any offending patterns across all of crates.io (I've also made it take visibility into account, which should match the version run against crater).

thomcc:
Yeah, I see. The issue isn't that code exists that would suddenly fail to compile, the issue is that this lint now exists and maybe we want to recompile existing functions to ensure they don't violate it.

I don't have strong feelings but I think that's a step too painful.

eeeebbbbrrrr (approved):

PR #322, opened 2023-06-02
Add a test that does a plrustc/rustc version check

thomcc:
Just checks that we can call both and that the versions match.

### DIFF
--- a/plrust/src/tests.rs
+++ b/plrust/src/tests.rs
@@ -1292,3 +1292,29 @@ tokio = { version = "1.19.2", features = ["rt", "net"]}"#
         ]
     }
 }
+
+#[test]
+#[cfg(feature = "trusted")]
+fn test_rustc_plrustc_version() {
+    fn rust_minor_version(rustc: &str) -> Option<u32> {
+        let output = std::process::Command::new(rustc)
+            .arg("--version")
+            .output()
+            .ok()?;
+        let version = std::str::from_utf8(&output.stdout).ok()?;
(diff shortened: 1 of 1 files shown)


plrust/src/tests.rs
```diff
@@ -1292,3 +1292,29 @@ tokio = { version = "1.19.2", features = ["rt", "net"]}"#
         ]
     }
 }
+
+#[test]
+#[cfg(feature = "trusted")]
```
eeeebbbbrrrr, line 1297: don't need to feature flag this.  If we merge this, it should happen regardless of trusted mode -- `plrustc` is used in both cases.

eeeebbbbrrrr (changes requested):

plrust/src/tests.rs
thomcc: Oh, is it?

eeeebbbbrrrr: Yeah.  We apply the lints no matter what.  `trusted` really just means "use postgrestd".

thomcc:
Yeah, I think a better solution here is to have plrust pass plrustc the sysroot explicitly, since it has all the information required to know it precisely.

This will have the added benefit of allowing me to finally kill the awful sysroot detection in plrustc: <https://github.com/tcdi/plrust/blob/9e5941eb9718d79bd133babfe25c25eeef0f4695/plrustc/plrustc/src/main.rs#L332-L381>, in addition to being more robust.

PR #329, opened 2023-06-09
A test for UDT/composite_type support

eeeebbbbrrrr:
The test is essentially a cut-n-paste of the long example in the docs.


BradyBonnette (approved):
I like it! (assuming CI does too)

Issue #325, opened 2023-06-06
Improve on using features of a dependency in plrust function


anth0nyleung:
PL/Rust does not handle using features for a dependencies quite well currently.

For instance, I have made serde an allowed dependency with the feature `derive`
```
$ > cat plrust_allowed_dependencies.toml
serde = { version = "1.0", features = ["derive"] }
```

But when i try to define a function with serde as dependency and enable the `derive` marco, it errors out because it expect the value of a dependency to be a string not a toml table
```
postgres=# CREATE FUNCTION foo (r INT, g INT, b INT) RETURNS INT AS
$$
CREATE FUNCTION table_deps (r INT, g INT, b INT) RETURNS TEXT STRICT AS
$$
                  [dependencies]
                  serde = { version = "1.0", features = ["derive"] }
                  rand = "*"
                  
                  [code]
                  use serde::{Serialize, Deserialize};
                  
                  #[derive(Debug, Serialize, Deserialize)]
                  struct Color {
                       r: u8,
                       g: u8,
                       b: u8,
                  }
                  
                  let r = r as u8;
                  let g = g as u8;
                  let b = b as u8;
                  
                  Ok(Some(format!("{:?}", Color{ r, g, b})))
                  $$ language plrust;        
$$ language plrust;

ERROR:
   0: dependency serde with values Table({"features": Array([String("derive")]), "version": String("1.0")}) is malformatted. Only strings are supported

Location:
   plrust/src/user_crate/mod.rs:355

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
postgres=#
```

eeeebbbbrrrr:
got a fix for this.  Simple oversight from #103.

Also note, I updated your example above so it compiles and executes with what I assume your intent was.  I didn't change it to specifically exercise serialization, other than keeping the `#[derive(Serialize, Deserialize)]`, which definitely engages the `serde` compile-time machinery!

eeeebbbbrrrr:
> got a fix for this. Simple oversight from https://github.com/tcdi/plrust/pull/103.

correction: it’s much more complex than I first thought. I think @thomcc ams I have a path forward tho.

eeeebbbbrrrr:
Resolved in v1.2.0.

PR #332, opened 2023-06-11
In response to issue #325, rework the dependency allow-list.

eeeebbbbrrrr:
Generally, the changes here allow a single dependency in the allow-list to have one **or more** versions specified, either as exact (`=`) or bounded ranges (`>=,<=`) or wildcards (`*`).

PL/Rust will pick the largest (newest, hopefully) entry from the allow-list that matches the function's requested version.  Depending on how either is written, the allow-list version could be used or the requested version could be used.

See the new documentation for more details.

---

As a drive-by, a number of our unit tests were tagged with `#[should_panic]` but without an expected message.  Turns out many of them were panicking but for the wrong reason!  This PR addresses these too.  

That said, the `plrust_pgloglevel_dont_allcaps_panic` test is currently set to `#[ignore]`.  Its intent is to block using Postgres' `PANIC` log level (which purposely crashes the backend), and I'm not yet clear on how this is supposed to be addressed.

### DIFF
--- a/plrust/src/allow_list.rs
+++ b/plrust/src/allow_list.rs
@@ -1,0 +1,434 @@
+use crate::gucs::PLRUST_ALLOWED_DEPENDENCIES;
+use semver::{BuildMetadata, Comparator, Op, Version, VersionReq};
+use std::cmp::Ordering;
+use std::collections::BTreeMap;
+use std::fmt::{Display, Formatter};
+use std::iter::once;
+use std::ops::Deref;
+use std::path::PathBuf;
+use std::str::FromStr;
+use toml::Value;
+
+#[derive(Debug, PartialEq)]
+pub struct Dependency {
+    name: String,
+    versions: BTreeMap<OrderedVersionReq, toml::value::Table>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+#[repr(transparent)]
+pub struct OrderedVersionReq(VersionReq);
+
+pub type AllowList = BTreeMap<String, Dependency>;
+
+impl Display for OrderedVersionReq {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl TryFrom<String> for OrderedVersionReq {
+    type Error = Error;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        OrderedVersionReq::try_from(value.as_str())
+    }
+}
+
+impl TryFrom<&str> for OrderedVersionReq {
+    type Error = Error;
+
+    /// only allow "*", exact, and bounded VersionReq values
+    fn try_from(version: &str) -> Result<Self, Self::Error> {
+        let vreq = VersionReq::parse(&version)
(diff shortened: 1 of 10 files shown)


plrust/src/allow_list.rs
```diff
@@ -0,0 +1,434 @@
...
+                    Error::MalformedVersion(wanted_version.to_string(), e.to_string())
+                })?;
+
+                for (version, table) in ranked_versions {
+                    if version.matches_versionreq(&wanted_version) {
+                        let mut table = table.clone();
+
+                        if version.to_string().contains('*') {
+                            // the matching allow-list VersionReq is a wildcard pattern, so since the
+                            // `wanted_version` matches it, use the (probably) more precise `wanted_version`
```
thomcc, line 313: Glad to see this only uses the version field from the user, since allowing the table might accidentally allow `path`/`git`/... deps (`path` deps allowing inspection of the machine, `git` deps allowing arbitrary code execution).

plrust/src/user_crate/mod.rs
```diff
@@ -352,103 +358,173 @@ fn parse_source_and_deps(
...
     for (dependency, val) in &user_dependencies {
         match val {
-            toml::Value::String(_) => {
-                // No-op, we currently only support dependencies in the format
-                // foo = "1.0.0"
-            }
+            // user dependencies in these general forms are allowed:
+            //    name = "x.y.z"
+            //    name = { version = "x.y.z", features = [ "a", "b", "c" ]
+            toml::Value::String(_) | toml::Value::Table(_) => {}
```
thomcc, line 369: Still TODO I'm guessing?

plrust/src/user_crate/mod.rs
```diff
@@ -352,103 +358,173 @@ fn parse_source_and_deps(
...
+        if let Some(wanted_table) = wanted_value.as_table() {
+            if wanted_table.len() > 1 {
+                let mut wanted = wanted_table.clone();
+                let mut used = used.as_table().unwrap().clone();
+
+                // don't compare the version values
+                wanted.remove("version");
+                used.remove("version");
+
+                if used != wanted {
```
thomcc, line 440: Great!

thomcc (commented):
What's here looks very good! I thought it was complete, but it looks like there's a little bit of WIP left?

plrust/src/allow_list.rs
```diff
@@ -38,26 +38,45 @@ impl TryFrom<String> for OrderedVersionReq {
...
             Err(Error::UnsupportedVersionReq(version.to_string()))
-        };
+        }
+    }
+}
+
+fn validate_versionreq(vreq: &VersionReq, require_exact: bool) -> bool {
+    let has_prelrease = vreq.comparators.iter().any(|cmp| !cmp.pre.is_empty());
+    if has_prelrease {
+        // -shitshow versions not allowed (https://docs.rs/semver/latest/semver/struct.Prerelease.html#examples)
```
thomcc, line 59: 💩

thomcc (approved):
Looks great.

andreas:
This change was really helpful by also allowing git dependencies -- thanks! 🙏

PR #334, opened 2023-06-12
Support `--features trusted` on macOS

eeeebbbbrrrr:
With this PR, PL/Rust now supports installation as a Trusted Procedural Language on macOS.

It also supports cross compiling between x86_64 and aarch64 on macOS.

We do **NOT** support cross-compiling between operating systems.


thomcc (approved):
Looks great. No objections -- the weird bits are non-public and this is much simpler than the alternative.

eeeebbbbrrrr:
Thanks for making CI work on macOS, @BradyBonnette

PR #335, opened 2023-06-13
Fully resolve aliases in plrust-suspicious-trait-object

thomcc:
What a hassle. The other lints aren't sensitive to this but just by coincidence -- using aliases in them causes the I-unsound Rust issue to resolve itself. I'll see if I can figure out a less error-prone way to do those.


eeeebbbbrrrr (approved):
LGTM.  Thanks for the offline discussion about this.

PR #341, opened 2023-06-21
correct docs and code comments around the `plrust.required_lints` GUC

eeeebbbbrrrr:
This just fixes some inaccuracies in the docs and code comments that were missed when we changed how `plrust.required_lints` is applied in #327


eeeebbbbrrrr:
I'd like to also go ahead and merge this into `main` too since it's really a doc update

eeeebbbbrrrr:
I cherry-picked this into `main`

PR #342, opened 2023-06-21
Significantly simplify sysroot configuration

thomcc:
We no longer have to manually search for the sysroot, as of https://github.com/rust-lang/rust/pull/103660 it should be auto-detected from first the arguments passed, followed by locating it based on the location of the rustc_driver dynamic library.

This means we don't support configuring it via the environment anymore, but I don't believe anybody was explicitly using this anyway. Regardless, if they were configuring it to a different value than we locate with this function, it was almost certainly wrong. Still, we should make a note in the next release notes about it.

Thanks to @jyn514 for pointing this out to me!

### DIFF
--- a/plrustc/plrustc/src/main.rs
+++ b/plrustc/plrustc/src/main.rs
@@ -19,8 +19,7 @@ use rustc_session::early_error;
 use rustc_session::parse::ParseSess;
 use rustc_span::source_map::FileLoader;
 use rustc_span::Symbol;
-use std::ffi::OsStr;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
 const PLRUSTC_USER_CRATE_NAME: &str = "PLRUSTC_USER_CRATE_NAME";
 const PLRUSTC_USER_CRATE_ALLOWED_SOURCE_PATHS: &str = "PLRUSTC_USER_CRATE_ALLOWED_SOURCE_PATHS";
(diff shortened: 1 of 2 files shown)


plrustc/plrustc/src/main.rs
```diff
@@ -108,45 +109,14 @@ fn main() {
...
-            .get(1)
-            .map(std::path::Path::new)
-            .and_then(std::path::Path::file_stem)
-            .map_or(false, |name| {
-                name == our_exe_filename || name == "plrustc" || name == "rustc"
-            });
-
-        if wrapper_mode {
-            args.remove(1);
-        }
```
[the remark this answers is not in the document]
thomcc: Yes, I don't think we realistically supported that anyway.

plrustc/plrustc/src/main.rs
```diff
@@ -52,6 +51,8 @@ impl Callbacks for PlrustcCallbacks {
     }
 }
 
+// TODO: eventually we can replace this with:
+// rustc_driver::install_ice_hook("https://github.com/tcdi/plrust/issues/new", |_| ());
```
workingjubilee, line 55: When? :eyes:

plrustc/plrustc/src/main.rs
```diff
@@ -108,45 +109,14 @@ fn main() {
...
-            .map_or(false, |name| {
-                name == our_exe_filename || name == "plrustc" || name == "rustc"
-            });
-
-        if wrapper_mode {
-            args.remove(1);
-        }
-
+        let args = rustc_driver::args::arg_expand_all(&std::env::args().collect::<Vec<_>>());
+        let config = PlrustcConfig::from_env_and_args(&args);
```
workingjubilee, line 113: So much less code!

workingjubilee (approved):
It's beautiful.

plrustc/plrustc/src/main.rs
thomcc: We actually don't use nightly >_>

jyn514: ah. when 1.71 lands on stable in a few weeks, then.

PR #343, opened 2023-06-23
Feike/toolchain fix

feikesteenbergen:
In commit 4f71dd52 the default version is bumped to 1.70.0, yet these
instructions still carry the old version.
Using 1.67.1 fails using the current latest release tag, wherease 1.70.0
succeeds.

### DIFF
--- a/doc/book.toml
+++ b/doc/book.toml
@@ -6,4 +6,4 @@ src = "src"
 title = "PL/Rust Guide"
 
 [preprocessor.variables.variables]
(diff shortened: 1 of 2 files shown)


eeeebbbbrrrr:
Hey, thanks @feikesteenbergen!  I'll merge this through to the `main` branch now so that the docs site will automatically update.  For the next release I'll also write a little document to list out the things we need to update when things change.

eeeebbbbrrrr (approved):

PR #346, opened 2023-06-30
Add a lint for closure trait impls

thomcc:
This sucked to write, but should be good enough. It will catch a lot of external crates incorrectly though, without the refinement.

### DIFF
--- a/plrustc/plrustc/src/lints/closure_trait_impl.rs
+++ b/plrustc/plrustc/src/lints/closure_trait_impl.rs
@@ -1,0 +1,81 @@
+//! Ideally we'd do this in such a way that it only forbids something that
+//! allows projection through to the return type, for example
+//! ```ignore (exposition-only)
+//! trait Trait {
+//!     type Assoc;
+//! }
+//! impl<R, F: Fn() -> R> Trait for F {
+//!     type Assoc = R;
+//! }
+//! ```
+//! allows `<_ as Trait>::Assoc` to get a functions return type. That said,
+//! actually writing this lint has totally defeated me at the moment, so this is
+//! good enough for now.
+use rustc_hir as hir;
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+
+declare_plrust_lint!(
+    pub(crate) PLRUST_CLOSURE_TRAIT_IMPL,
+    "Disallow trait impls which are generic over closure type",
+);
+
+rustc_lint_defs::declare_lint_pass!(PlrustClosureTraitImpl => [PLRUST_CLOSURE_TRAIT_IMPL]);
+
+impl<'tcx> LateLintPass<'tcx> for PlrustClosureTraitImpl {
+    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx hir::Item<'tcx>) {
+        let hir::ItemKind::Impl(impl_item) = item.kind else {
+            return
+        };
+        for pred in impl_item.generics.predicates {
+            let hir::WherePredicate::BoundPredicate(bound_pred) = pred else {
+                continue
+            };
+            // TODO: should we ignore cases where `bound_pred.bounded_ty` isn't
+            // from from one of `item.generics.params`?
+            for bound in bound_pred.bounds {
+                match bound {
+                    hir::GenericBound::LangItemTrait(
+                        hir::LangItem::Fn
+                        | hir::LangItem::FnOnce
+                        | hir::LangItem::FnMut
+                        | hir::LangItem::FnPtrTrait,
+                        ..,
+                    ) => {
+                        cx.lint(
+                            PLRUST_CLOSURE_TRAIT_IMPL,
+                            "trait impls bounded on function traits are forbidden in PL/Rust",
+                            |b| b.set_span(bound_pred.span),
+                        );
+                    }
+                    hir::GenericBound::LangItemTrait(..) => {
+                        // Don't care about other traits (I think)
+                    }
+                    hir::GenericBound::Trait(poly_trait, ..) => {
+                        let Some(impl_did) = poly_trait.trait_ref.path.res.opt_def_id() else {
+                            continue
+                        };
+                        let lang_items = cx.tcx.lang_items();
+                        let fntraits = [
+                            lang_items.get(hir::LangItem::Fn),
+                            lang_items.get(hir::LangItem::FnOnce),
+                            lang_items.get(hir::LangItem::FnMut),
+                            lang_items.get(hir::LangItem::FnPtrTrait),
+                        ];
+                        if fntraits.contains(&Some(impl_did)) {
+                            cx.lint(
+                                PLRUST_CLOSURE_TRAIT_IMPL,
+                                "trait impls bounded on function traits are forbidden in PL/Rust",
+                                |b| b.set_span(bound_pred.span),
+                            );
+                        }
+                        // TODO: if that fails, do we need to
+                        // try_normalize_erasing_regions and retry?
+                    }
+                    hir::GenericBound::Outlives(..) => {
+                        // Don't care about these.
+                    }
+                }
+            }
+        }
+    }
+}
--- a/plrustc/plrustc/src/lints/closure_trait_impl.rs
+++ b/plrustc/plrustc/src/lints/closure_trait_impl.rs
@@ -11,6 +11,19 @@
 //! allows `<_ as Trait>::Assoc` to get a functions return type. That said,
 //! actually writing this lint has totally defeated me at the moment, so this is
 //! good enough for now.
+//!
+//! For some intuition as to why this is tricky, consider cases like
+//! ```ignore (exposition-only)
+//! trait GivesAssoc<A> { type Assoc; }
+//! impl<A> GivesAssoc<A> for A { type Assoc = A; }
+//!
+//! trait Child<T> where Self: GivesAssoc<T> {}
+//! impl<R, F: Fn() -> R> Child<R> for F {}
+//! ```
+//! and similarly complicated variants. To figure this out you need to examine
+//! not just the directly implemented trait, but also all traits that are
+//! indirectly implemented via bounds as a result of the impl. This is possible,
+//! but... difficult.
 use rustc_hir as hir;
 use rustc_lint::{LateContext, LateLintPass, LintContext};
 
--- a/plrustc/plrustc/src/lints/closure_trait_impl.rs
+++ b/plrustc/plrustc/src/lints/closure_trait_impl.rs
@@ -44,7 +44,7 @@ impl<'tcx> LateLintPass<'tcx> for PlrustClosureTraitImpl {
                 continue
             };
             // TODO: should we ignore cases where `bound_pred.bounded_ty` isn't
-            // from from one of `item.generics.params`?
+            // from one of `item.generics.params`?
             for bound in bound_pred.bounds {
                 match bound {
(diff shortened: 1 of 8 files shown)


plrustc/plrustc/src/lints/closure_trait_impl.rs
```diff
@@ -0,0 +1,81 @@
...
+        };
+        for pred in impl_item.generics.predicates {
+            let hir::WherePredicate::BoundPredicate(bound_pred) = pred else {
+                continue
+            };
+            // TODO: should we ignore cases where `bound_pred.bounded_ty` isn't
+            // from from one of `item.generics.params`?
+            for bound in bound_pred.bounds {
+                match bound {
+                    hir::GenericBound::LangItemTrait(
```
thomcc, line 37: Only checking this doesn't seem to work reliably, for some reason.

plrustc/plrustc/src/lints/closure_trait_impl.rs
```diff
@@ -0,0 +1,84 @@
...
+//! actually writing this lint has totally defeated me at the moment, so this is
+//! good enough for now.
+//!
+//! For some intuition as to why this is tricky, consider cases like
+//! ```ignore (exposition-only)
+//! trait GivesAssoc<A> { type Assoc; }
+//! impl<A> GivesAssoc<A> for A { type Assoc = A; }
+//!
+//! trait Child<T> where Self: GivesAssoc<T> {}
+//! impl<R, F: Fn() -> R> Child<R> for F {}
```
workingjubilee, line 21: Ah, so we have to be alert to implementations that "smuggle" in type projections, even though the trait we're implementing does not allow type projection, because the trait implemented then allows us to find another trait which allows the type projection.

plrustc/plrustc/src/lints/closure_trait_impl.rs
```diff
@@ -0,0 +1,84 @@
...
+//! ```
+//! and similarly complicated variants. To figure this out you need to examine
+//! not just the directly implemented trait, but also all traits that are
+//! indirectly implemented via bounds as a result of the impl. This is possible,
+//! but... difficult.
```
workingjubilee, line 26: So we're just banhammering all of them on the function types, right?

plrustc/plrustc/src/lints/fn_ptr.rs
```diff
@@ -22,22 +22,13 @@ impl<'tcx> LateLintPass<'tcx> for PlrustFnPointer {
...
-                                    "Use of function trait objects is forbidden in PL/Rust",
-                                    |b| b.set_span(ty.span),
-                                );
-                            }
-                        }
+                for poly_trait in *traits {
+                    if super::utils::has_fn_trait(cx, poly_trait) {
+                        cx.lint(
+                            PLRUST_FN_POINTERS,
+                            "Use of function trait objects is forbidden in PL/Rust",
```
workingjubilee, line 29: Even primitive function pointers? Confused about the naming, if not.

plrustc/plrustc/src/lints/closure_trait_impl.rs
```diff
@@ -0,0 +1,84 @@
...
+                    hir::GenericBound::Trait(poly_trait, ..) => {
+                        if super::utils::has_fn_trait(cx, poly_trait) {
+                            cx.lint(
+                                PLRUST_CLOSURE_TRAIT_IMPL,
+                                "trait impls bounded on function traits are forbidden in PL/Rust",
+                                |b| b.set_span(bound_pred.span),
+                            );
+                        }
+                        // TODO: if that fails, do we need to
+                        // try_normalize_erasing_regions and retry?
```
workingjubilee, line 75: Do we?

plrustc/plrustc/src/lints/closure_trait_impl.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx hir::Item<'tcx>) {
+        let hir::ItemKind::Impl(impl_item) = item.kind else {
+            return
+        };
+        for pred in impl_item.generics.predicates {
+            let hir::WherePredicate::BoundPredicate(bound_pred) = pred else {
+                continue
+            };
+            // TODO: should we ignore cases where `bound_pred.bounded_ty` isn't
+            // from from one of `item.generics.params`?
```
workingjubilee, line 47: I don't know right away, but
```suggestion
            // TODO: should we ignore cases where `bound_pred.bounded_ty` isn't
            // from one of `item.generics.params`?
```

plrustc/plrustc/src/lints/closure_trait_impl.rs
```diff
@@ -0,0 +1,84 @@
...
+                                PLRUST_CLOSURE_TRAIT_IMPL,
+                                "trait impls bounded on function traits are forbidden in PL/Rust",
+                                |b| b.set_span(bound_pred.span),
+                            );
+                        }
+                        // TODO: if that fails, do we need to
+                        // try_normalize_erasing_regions and retry?
+                    }
+                    hir::GenericBound::Outlives(..) => {
+                        // Don't care about these.
```
workingjubilee, line 78: More specific, please
```suggestion
                        // Lifetime generics are irrelevant for guards against type projection
```
or no comment at all is fine.

plrustc/plrustc/src/lints/closure_trait_impl.rs
workingjubilee: presumably we aren't concerned about Sized or Copy?

plrustc/plrustc/src/lints/closure_trait_impl.rs
```diff
@@ -0,0 +1,84 @@
...
+                        ..,
+                    ) => {
+                        cx.lint(
+                            PLRUST_CLOSURE_TRAIT_IMPL,
+                            "trait impls bounded on function traits are forbidden in PL/Rust",
+                            |b| b.set_span(bound_pred.span),
+                        );
+                    }
+                    hir::GenericBound::LangItemTrait(..) => {
+                        // Don't care about other traits (I think)
```
workingjubilee, line 64: Probably should be more like
```suggestion
                        //  other stable builtin traits aren't useful for projecting a function's return type
```

workingjubilee (commented):
Either `closure_trait_impl.rs` or `config-lints.md` should explain the specific problem more exactly: We want to ban projecting a return type that is incorrectly considered `'static` despite in actuality being lifetime-bound.

plrustc/plrustc/src/lints/fn_ptr.rs
thomcc: Primitive function pointers are handled above. I don't think you can call through `FnPtr` anyway.

plrustc/plrustc/src/lints/closure_trait_impl.rs
thomcc: I will take your suggestion. I prefer having some comment to none to indicate "yes I thought about this case and am deliberately leaving it out". But providing rationale is better.

plrustc/plrustc/src/lints/closure_trait_impl.rs
thomcc: Not as far as I know. I couldn't make it work anyway. I think this is where we'd do it if we were going to put it somewhere, though!

We probably should do this for the `plrust_fn_pointers` lint, but doing so would likely break way too much code that could only be unsound in the presence of malicious (or extremely strangely written) external crates.

plrustc/plrustc/src/lints/closure_trait_impl.rs
thomcc: Yep, not concerned at all.

plrustc/plrustc/src/lints/closure_trait_impl.rs
thomcc: Yup.

plrustc/plrustc/src/lints/closure_trait_impl.rs
thomcc: Basically.

plrustc/plrustc/src/lints/fn_ptr.rs
workingjubilee: Ah right. Carry on!

plrustc/plrustc/src/lints/closure_trait_impl.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx hir::Item<'tcx>) {
+        let hir::ItemKind::Impl(impl_item) = item.kind else {
+            return
+        };
+        for pred in impl_item.generics.predicates {
+            let hir::WherePredicate::BoundPredicate(bound_pred) = pred else {
+                continue
+            };
+            // TODO: should we ignore cases where `bound_pred.bounded_ty` isn't
+            // from one of `item.generics.params`?
```
workingjubilee, line 47: Why would it be sound to ignore them?

plrustc/plrustc/src/lints/closure_trait_impl.rs
workingjubilee: We must imagine Sisyphus LOLing.

plrustc/plrustc/src/lints/closure_trait_impl.rs
thomcc: Because I think they would have to be concrete. E.g. I don't believe `impl<F> Foo for Blah where SomethingElse: Fn()` is problematic.

thomcc: But also that's such a pathological case that I'm not worried about it; the only way I could see it happening is from the output of some macro, and for that I'd want to see the actual code.

That is, if someone files a bug we could consider it. I think it would be sound, but it would complicate this code quite a bit.

workingjubilee: Fair enough, then.

workingjubilee:
It would be nice to have an example of a crate that this too-conservative lint would nail that we want to allow but I consider that purely extra. This will be good with the doc change I mentioned.

workingjubilee (approved):
With this diff I mentioned wanting, or some variation of this:

thomcc:
> It would be nice to have an example of a crate that this too-conservative lint would nail that we want to allow but I consider that purely extra

I don't think this is what you mean, but for clarity: if you are asking for this as a uitest, I do not want to start adding known-bug uitests yet, I think it could cause confusion, and I would eventually like to add some rules that we should be careful about reviewing changes to the stderr uitest files, since they currently should be very rare.

If you just want an example, well, it's still tricky but only because the pattern doesnt lend itself to github code search. These impls are semi-common, because the function traits are `#[fundamental]` -- the blanket impl doesn't forbid other things. As mentioned it's hard to search for, but <https://github.com/rust-lang/rust/blob/fd68a6ded951bd7b852ab8107007f7145e3ad6ec/library/core/src/str/pattern.rs#L602-L610> is an example from the stdlib (and I'm sure there are more).

Such impls are fine, because they don't give access to the return type as an associated type. This is true if they show up in dependencies, this isn't a case where we are weakened if a dep is added where this falsely triggers (although if a dep did provide access to the return type, it would be a problem).

workingjubilee:
definitely not as a uitest, just wondering if we can find examples to use to help motivate our own discussions/research later. I expect we'll encounter them face-first tho'.

PR #357, opened 2023-07-13
Drop old tempdir dep

workingjubilee:
tempfile provides functions for tempdirs too,
so we do not need an old, deprecated dep.
This also gets rid of a minor security advisory,
not actually exploitable in our usage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -512,12 +512,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


thomcc (approved):

workingjubilee:
@workingjubilee You have to actually run `cargo test` in order to check if tests compile, Jubilee, you can't get away with a quick `cargo check` unless it is `cargo check --tests`.

PR #360, opened 2023-07-18
Fixes documentation generation CI

BradyBonnette:
There was some kind of incompatibility issue with the latest version of `mdbook` and `mdbook-variables`, so these changes use specific versions of both crates that are known to work. We do not need the bleeding edge of either crate at the moment.

NOTE: This is going right into `tcdi/plrust:main` since this is only a CI change, and the changes won't come

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,342 +1,342 @@
-# name: PL/Rust
-
-# on:
-#   push:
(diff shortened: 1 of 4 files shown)


eeeebbbbrrrr (approved):
Thank you @BradyBonnette!  LGTM, merging now

Issue #371, opened 2023-08-07
Add a function to list all the allowlisted crates with their respective versions, features


anth0nyleung:
Since we have created introduced the mechanism to support allow-listing certain crates, it'd be useful to have a function that lists what are the allow-listed crates with their respective versions, and features etc.

One idea is to make a srf like.

```
postgres=# select plrust.allowed_dependencies();
 allowed_dependencies 
----------------------
 (serde,1.0,{derive})
(1 row)

postgres=# 
```

```
$ > cat ~/allow.toml 
serde = { version = "=1.0", feature = ["derive"] }
```

eeeebbbbrrrr:
This is a fine idea. Are you planning on submitting a PR or would you prefer we do it?

anth0nyleung:
I would be happy to work on it!

PR #373, opened 2023-08-11
Add a function to list all the allowlisted crates with their respecti…

anth0nyleung:
This commit addresses issue #371 

Introduce a function to list all the allowed dependencies and their respective versions and features.

E.g Given a toml like this,
```
$ > cat ~/allow.toml
owo-colors = "=3.5.0"
tokio = { version = "=1.19.2", features = ["rt", "net"] }
plutonium = "*"
syn = { version = "=2.0.28", default-features = false }
rand = ["=0.8.3", { version = ">0.8.4, <0.8.6", features = ["getrandom"] }]
```
the function returns a table
```
postgres=# SELECT * FROM plrust.list_allowed_dependencies();
    name    |    version     |  features   | default_features
------------+----------------+-------------+------------------
 owo-colors | =3.5.0         | {}          | t
 plutonium  | *              | {}          | t
 rand       | =0.8.3         | {}          | t
 rand       | >0.8.4, <0.8.6 | {getrandom} | t
 syn        | =2.0.28        | {}          | f
 tokio      | =1.19.2        | {rt,net}    | t
(6 rows)
```

This commit also bumps plrust to 1.1 and introduce the sql script for schema upgrade

---

### Testing

- Unit test passed
- Tested extension upgrade

1. Install plrust 1.0 (v1.2.3)
```
postgres=# CREATE EXTENSION plrust ;
CREATE EXTENSION
postgres=# \dx plrust
List of installed extensions
-[ RECORD 1 ]-----------------------------------------------------------
Name        | plrust
Version     | 1.0
Schema      | plrust
Description | plrust:  A Trusted Rust procedural language for PostgreSQL

postgres=# \df plrust.*
List of functions
-[ RECORD 1 ]-------+--------------------
Schema              | plrust
Name                | plrust_call_handler
Result data type    | language_handler
Argument data types |
Type                | func
-[ RECORD 2 ]-------+--------------------
Schema              | plrust
Name                | plrust_validator
Result data type    | void
Argument data types | fn_oid oid
Type                | func
```
2. Update plrust to 1.1, new function is installed
```
postgres=# ALTER EXTENSION plrust UPDATE TO '1.1';
ALTER EXTENSION
postgres=# \dx plrust
List of installed extensions
-[ RECORD 1 ]-----------------------------------------------------------
Name        | plrust
Version     | 1.1
Schema      | plrust
Description | plrust:  A Trusted Rust procedural language for PostgreSQL

postgres=# \df plrust.*
List of functions
-[ RECORD 1 ]-------+--------------------------------------------------------------------------
Schema              | plrust
Name                | list_allowed_dependencies
Result data type    | TABLE(name text, version text, features text[], default_features boolean)
Argument data types |
Type                | func
-[ RECORD 2 ]-------+--------------------------------------------------------------------------
Schema              | plrust
Name                | plrust_call_handler
Result data type    | language_handler
Argument data types |
Type                | func
-[ RECORD 3 ]-------+--------------------------------------------------------------------------
Schema              | plrust
Name                | plrust_validator
Result data type    | void
Argument data types | fn_oid oid
Type                | func

postgres=# SELECT * FROM plrust.list_allowed_dependencies();
    name    |    version     |  features   | default_features
------------+----------------+-------------+------------------
 owo-colors | =3.5.0         | {}          | t
 plutonium  | *              | {}          | t
 rand       | =0.8.3         | {}          | t
 rand       | >0.8.4, <0.8.6 | {getrandom} | t
 syn        | =2.0.28        | {}          | f
 tokio      | =1.19.2        | {rt,net}    | t
(6 rows)
```


eeeebbbbrrrr:
Thanks!  I’ll provide some feedback in the morning.

plrust/src/allow_list.rs
```diff
@@ -212,6 +212,53 @@ pub enum Error {
...
+                    .collect(),
+                false => vec![features.as_str().unwrap().to_owned()],
+            },
+            None => vec![],
+        };
+        let default_features = match version.get("default-features") {
+            Some(v) => v.as_bool().unwrap(),
+            None => true,
+        };
+        tuples.push((name, ver, features, default_features));
```
workingjubilee, line 257: Due to field punning, this doesn't become much more verbose when using the struct initialization syntax:
```suggestion
        tuples.push(AllowedDependency { name, ver, features, default_features });
```

plrust/plrust.control
```diff
@@ -1,6 +1,6 @@
 comment = 'plrust:  A Trusted Rust procedural language for PostgreSQL'
-default_version = '1.0'
+default_version = '1.1'
 module_pathname = '$libdir/plrust'
 relocatable = false
 superuser = false
-schema = plrust
+schema = plrust
```
workingjubilee, line 6: ?
```suggestion
schema = plrust

```

plrust/src/tests.rs
```diff
@@ -1398,6 +1398,74 @@ insert into people (p) values (make_person('Dr. Beverly Crusher of the Starship
...
+        //  owo-colors | =3.5.0         | {}          | t
+        //  plutonium  | *              | {}          | t
+        //  rand       | =0.8.3         | {}          | t
+        //  rand       | >0.8.4, <0.8.6 | {getrandom} | t
+        //  syn        | =2.0.28        | {}          | f
+        //  tokio      | =1.19.2        | {rt,net}    | t
+
+        Spi::connect(|client| {
+            let expected_names = vec!["owo-colors", "plutonium", "rand", "rand", "syn", "tokio"];
+            let expected_veresions = vec![
```
workingjubilee, line 1424:
```suggestion
            let expected_versions = vec![
```

plrust/src/tests.rs
```diff
@@ -1398,6 +1398,74 @@ insert into people (p) values (make_person('Dr. Beverly Crusher of the Starship
...
+            assert_eq!(tup_table.len(), expected_len);
+
+            for (i, row) in tup_table.into_iter().enumerate() {
+                assert_eq!(
+                    row["name"].value::<String>().unwrap(),
+                    Some(expected_names[i].to_owned())
+                );
+                assert_eq!(
+                    row["version"].value::<String>().unwrap(),
+                    Some(expected_veresions[i].to_owned())
```
workingjubilee, line 1454:
```suggestion
                    Some(expected_versions[i].to_owned())
```

plrust/src/lib.rs
```diff
@@ -193,6 +193,25 @@ unsafe fn plrust_call_handler(fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum
...
+                name!(features, Vec<String>),
+                name!(default_features, bool),
+            ),
+        >,
+    >,
+    Box<dyn std::error::Error + Send + Sync + 'static>,
+> {
+    let allowed_dependencies = allow_list::get_allowed_dependencies_as_tuple_vector();
+    Ok(Some(TableIterator::new(allowed_dependencies)))
+}
```
workingjubilee, line 213: The only unfortunate thing is I don't think there's an easy way yet to simply yield a TableIterator that yields the equivalent of a Rust struct so I guess this would have to `.map()` the struct back down to a tuple anyway... hmm... annoying... I should fix that. I mean, PL/Rust users aren't exposed to how grody it is, but

plrust/src/allow_list.rs
```diff
@@ -212,6 +212,53 @@ pub enum Error {
...
+    }
+    tuples
+}
```
workingjubilee, line 260: Not required: this entire thing performs pure iteration and allocates to do so, but it could be entirely a single

```rust
dep.versions.values().map(|version| {
    /* for loop's inner code here, except instead of pushing,
     * yielding each element from the mapping expression
     */
    AllowedDependency { name, ver, features, default_features }
})
```

and then return a signature of `impl Iterator<Item = AllowedDependency>`.

I say "not required" because this is one of those things that you may find easy to do orrr mildly infuriating.

plrust/src/allow_list.rs
```diff
@@ -212,6 +212,53 @@ pub enum Error {
...
+
+// Convert all the allowed dependencies into a vector of tuples
+pub fn get_allowed_dependencies_as_tuple_vector() -> Vec<AllowedDependencyTuple> {
+    let allowlist = load_allowlist().expect("Error loading dependency allow-list");
+    let mut tuples: Vec<AllowedDependencyTuple> = vec![];
+
+    for dependency in allowlist.values() {
+        let mut tuple: Vec<AllowedDependencyTuple> =
+            get_single_dependency_as_tuple_vector(dependency);
+        tuples.append(&mut tuple);
```
workingjubilee, line 226: This could consume the individual tuple instead of using `.append` to move out of it, since we don't use it after this and moving-in is thus legal, using `std::iter::Extend`. This would also allow using anything that returns a correct type when iterated over as the return from `get_single_dependency`, which would include both Vecs and other things.

```suggestion
        tuples.extend(tuple);
```

workingjubilee (commented):
Some nits.

plrust/src/allow_list.rs
```diff
@@ -212,6 +212,53 @@ pub enum Error {
...
+}
+
+// Multiple version of a dependencies can be allowed. Create a tuple for each version of the dependency
+fn get_single_dependency_as_tuple_vector(dep: &Dependency) -> Vec<AllowedDependencyTuple> {
+    let mut tuples: Vec<AllowedDependencyTuple> = vec![];
+    for version in dep.versions.values() {
+        let name = dep.name.clone();
+        let ver = version
+            .get("version")
+            .unwrap()
```
eeeebbbbrrrr, line 238: If it's possible for the "version" to be None, then we should account for that here and in our data structure instead of panicking.

plrust/src/allow_list.rs
```diff
@@ -212,6 +212,53 @@ pub enum Error {
...
+            .to_string()
+            .replace("\"", "");
+        let features: Vec<String> = match version.get("features") {
+            Some(features) => match features.is_array() {
+                true => features
+                    .as_array()
+                    .unwrap()
```
eeeebbbbrrrr, line 245: Same here, if it's possible that "features" won't be an array, then we should for this situation in the data structure instead of panicking.

plrust/src/allow_list.rs
```diff
@@ -212,6 +212,53 @@ pub enum Error {
     UnsupportedVersionReq(String),
 }
 
+// Tuple describes name of the dependency, version of the dependency, features of the dependecy that are enabled, whether default features enabled
+pub type AllowedDependencyTuple = (String, String, Vec<String>, bool);
```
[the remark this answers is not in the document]
eeeebbbbrrrr: I agree with Jubilee here.  This should be a struct with named members.  In addition to Jubilee's points, a proper type also avoid confusion around which tuple element means what.

plrust/src/lib.rs
```diff
@@ -193,6 +193,25 @@ unsafe fn plrust_call_handler(fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum
     }
 }
 
+#[pg_extern]
+fn list_allowed_dependencies<'a>() -> Result<
```
eeeebbbbrrrr, line 197: I think I'd prefer the public-facing function simply be named `allowed_dependencies`.  It's just a style thing, I suppose, but I prefer function names describe what they return not what they do.

eeeebbbbrrrr (changes requested):

plrust/src/allow_list.rs
anth0nyleung: If i understand correctly it should not panic here, because before calling `unwrap`,  i have already first checked if the field `features` exists, and if so whether the value for the field `feature` is an array, only then i would unwrap the value.

plrust/src/allow_list.rs
anth0nyleung: `version` can be None, but i think `load_allowlist` is already checking if `version` is set for an allowed dependency

```
(23-08-17 12:55:17) [~/pg_data]
$ > cat ~/allow.toml
foo = { features = ["bar"] }

(23-08-17 12:55:58) [~/pg_data]
$> psql -d postgres -c 'SELECT * FROM plrust.allowed_dependencies();'
ERROR:  Error loading dependency allow-list:
   0: Dependency entry is missing the `version` attribute

Location:
   plrust/src/allow_list.rs:432

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.

File: plrust/src/allow_list.rs [hunk 12]
```diff
@@ -212,6 +212,92 @@ pub enum Error {
...
+                    item.features.to_owned(),
+                    item.default_features,
+                )
+            })
+            .collect()
+    }
+}
+
+/// Get all the allowed dependencies entries as a AllowedDependencies struct
+pub fn get_alllowed_dependencies() -> AllowedDependencies {
```
workingjubilee, line 249:
```suggestion
pub fn get_allowed_dependencies() -> AllowedDependencies {
```

plrust/src/allow_list.rs
workingjubilee: Yeah, I think we're fine here. A dependency is only coherent with a version of some sort, since a Rust dependency, as far as cargo is concerned, is not a crate name but also its version, so panicking  is okay.

plrust/src/allow_list.rs
workingjubilee: Actually, the code should simply use `as_array().unwrap()` instead of first checking `is_array()`, because if there is an entry, it is only valid for it to be an array.

plrust/src/allow_list.rs
```diff
@@ -212,6 +212,92 @@ pub enum Error {
...
+            },
+            None => vec![],
+        };
+        let default_features = match version.get("default-features") {
+            Some(default_features) => match default_features.is_bool() {
+                true => default_features.as_bool().unwrap(),
+                false => false,
+            },
+            None => true,
+        };
```
workingjubilee, line 290: Same concern as above, only one possible type is acceptable:

```suggestion
        let default_features = version
            .get("default-features")
            .map(|b| b.as_bool().unwrap())
            .unwrap_or(true);
```

plrust/src/tests.rs
```diff
@@ -1426,7 +1494,8 @@ pub mod pg_test {
 owo-colors = "=3.5.0"
 tokio = { version = "=1.19.2", features = ["rt", "net"]}
 plutonium = "*"
-rand = "*"
+syn = { version = "=2.0.28", default-features = false }
+rand = ["=0.8.3", { version = ">0.8.4, <0.8.6", features = ["getrandom"]}]
```
workingjubilee, line 1498: Hmm, do we... actually want this? Is this valid in a Cargo.toml? I would think not?

plrust/src/lib.rs
```diff
@@ -193,6 +194,26 @@ unsafe fn plrust_call_handler(fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum
...
+                name!(version, String),
+                name!(features, Vec<String>),
+                name!(default_features, bool),
+            ),
+        >,
+    >,
+    Box<dyn std::error::Error + Send + Sync + 'static>,
+> {
+    let allowed_dependencies: Vec<AllowedDependencyTuple> =
+        allow_list::get_alllowed_dependencies().into();
```
workingjubilee, line 213:
```suggestion
        allow_list::get_allowed_dependencies().into();
```

workingjubilee (commented):
Apologies, Eric's remarks highlighted some things I missed on my first scan.

plrust/src/tests.rs
anth0nyleung: this is not a valid Cargo.toml format, but i think plrust is able to interpret that format.

https://tcdi.github.io/plrust/dependencies.html#the-format

plrust/src/allow_list.rs
anth0nyleung: yea ideally it should be an array, but apparently i think the code that parses the toml file does not check if the value is an array of not

https://github.com/tcdi/plrust/blob/main/plrust/src/allow_list.rs#L218

plrust/src/allow_list.rs
anth0nyleung: similar as above, the code that parses the file does not check the type of the value, so i assume `false` if the value is not a bool. Maybe that can be fixed separately.

plrust/src/allow_list.rs
anth0nyleung: Actually on second thoughts, i have added some code to validate that features is any array of string

plrust/src/allow_list.rs
anth0nyleung: I have added some code to validate that default-features is a boolean

plrust/src/tests.rs
workingjubilee: I... missed that in my code review after my vacation.

plrust/src/allow_list.rs
workingjubilee: Yes, I am fine with addressing the larger concern re: that file in another PR. I just think we should be locking that down. The entries should more-or-less match what a Cargo.toml accepts, to minimize the principle of surprise.

workingjubilee: Ah, I see you did some of the changes already. Thank you!

plrust/src/tests.rs
workingjubilee: Ah okay I see what's going on here at second glance. This is an array of entries for a single version, and each is actually independent. It first struck me as a bit weird but I grok the intention here now.

workingjubilee (approved):
This looks good to me, thank you!

PR #380, opened 2023-09-07
Fixes MacOS nightly; Introduces nightly mdbook build

BradyBonnette:
Nightly MacOS test runs were failing due to the updated integration test structure recently introduced to the project. Part of these changes addresses that situation.

The other part of the changes introduces an mdbook nightly build. This does not publish the documentation, it only checks to make sure that mdbook+plugins can install and build the documentation properly. I believe it was necessary because the only time documentation is built is when a release happens, and if documentation generation fails for whatever reason, we find out only after a release has been published. The hope is that this should catch any errors before we release.

NOTE: This PR wants to be merged into tcdi/plrust:main directly, as it is only CI changes and the nightly updates can be applied immediately instead of waiting for another plrust release.


eeeebbbbrrrr (approved):
I guess this is fine.  I'll merge now and we'll see what CI does with it.

Issue #377, opened 2023-09-06
fix `plrust/build` to dynamically determine pgrx version


eeeebbbbrrrr:
https://github.com/tcdi/plrust/blob/9f0f459899b8982623ae2b7b7ddbfcf394f87cb8/plrust/build#L56

That line needs to not hardcode the pgrx version.  The ideal solution is that it can figure it out from the sibling Cargo.toml file.

BradyBonnette:
Closing as a result of https://github.com/tcdi/plrust/pull/381 getting merged into `develop`

PR #381, opened 2023-09-07
Updates plrust/build to programmatically get pgrx version

BradyBonnette:
Addresses: https://github.com/tcdi/plrust/issues/377

When plrust goes through a release cycle, occasionally (read: pretty much "almost" every time) we forget to update the hard-coded pgrx version in the `plrust/build` script due to the fact that our automated release tooling does not consider nor modify the script file. With the changes in this PR, we programmatically get the pgrx version using `cargo tree` and some unix utils magic so that nobody has to think about updating that value ever again upon release.

### DIFF
--- a/plrust/build
+++ b/plrust/build
@@ -53,7 +53,7 @@ cargo update -p pgrx
 cargo fetch
 if [ "$CI" != true ]; then
     # Attempt to get pgrx version from cargo tree. Looking for a pattern such as "pgrx v0.10.0"
-    PGRX_VERSION=$(cargo tree | grep -E "\s*pgrx\s+v[0-9]+\.[0-9]+\.[0-9]+" | head -n 1 | cut -f2- -dv)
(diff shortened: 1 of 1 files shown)


plrust/build
```diff
@@ -52,9 +52,18 @@ fi
 cargo update -p pgrx
 cargo fetch
 if [ "$CI" != true ]; then
-    cargo install cargo-pgrx \
-     --version "0.10.0" \
-     --locked # use the Cargo.lock in the pgrx repo
+    # Attempt to get pgrx version from cargo tree. Looking for a pattern such as "pgrx v0.10.0"
+    PGRX_VERSION=$(cargo tree | grep -E "\s*pgrx\s+v[0-9]+\.[0-9]+\.[0-9]+" | head -n 1 | cut -f2- -dv)
```
BradyBonnette, line 56: Just realized that this should maybe be `cargo tree --package plrust`. Thoughts?

eeeebbbbrrrr: yeah, probably.  I think the thing I send you on Discord assumed being in the `plrust/` crate directory.  Testing manually it doesn't seem to matter tho.

Could also add `--depth 1`

eeeebbbbrrrr (approved):
My comment aside, I think this is fine.  If you're done done done, feel free to merge it.

PR #385, opened 2023-09-18
Migrates aarch64 CI runners to Ubuntu

BradyBonnette:
Ubuntu generally does not require workarounds or tricks to get Postgres + PL/Rust to play nicely, so we are going to move the CI runners to use Ubuntu 22.04 instead of AL2. This should make future updates a little less painless.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -261,7 +261,7 @@ jobs:
         echo ""
 
         echo "----- Remove old postgres and libraries -----"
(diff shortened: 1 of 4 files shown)


BradyBonnette:
Even though this is a CI change only, I'm putting it into `develop` because it makes more sense to get it working with current in-development branches. Once this is merged, this will ultimately need to be merged into any outstanding in-development branches.

eeeebbbbrrrr (approved):

PR #390, opened 2023-10-18
Add dynamic function call support

eeeebbbbrrrr:
This adds support for pgrx' `fn_call` interface that enables calling other schema-defined functions dynamically.

The only code changes, as it were, are in the `plrust-trusted-pgrx` reexport crate.  The bulk of this PR is adding a little chapter to the book along with a simple unit test.


BradyBonnette (approved):
Great documentation updates!

eeeebbbbrrrr:
The two CI failures are on pg16 x86 b/c of that stupid simd stuff pg16 introduced.  Today isn't the day to play whackamole to fix up CI.

PR #394, opened 2023-11-01
Clear env in user function

### LINKED ISSUE #393: env/option_env failing to be linted by plrustc
Because this is unrelated to https://github.com/tcdi/plrust/issues/387 it deserves it's own issue.

Fixing it outright seems difficult (still not sure how), I think it'd probably be fine to try wiping the env on entry to the plrust user crate though. Because it seems specific to the env macro (and not a more general thing) that should be enough.

thomcc:
Fixes #393, although changes the semantics to "we catch this with a lint" to "we ensure all environment vars are missing".

### DIFF
--- a/plrust-tests/src/trusted.rs
+++ b/plrust-tests/src/trusted.rs
@@ -190,7 +190,7 @@ mod tests {
 
     #[pg_test]
     #[search_path(@extschema@)]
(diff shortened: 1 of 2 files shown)


thomcc:
Hm, not quite.

